PR #2, opened 2019-05-17
Add questionnaire id

bitdivision:
This PR adds the questionnaire id to the claims to launch eq with and removes the case id.

It is to be used to test the following PR: https://github.com/ONSdigital/eq-survey-runner/pull/2084

https://trello.com/c/0fkw3X5F/2984-jwt-claims-questionnaireid-and-caseid

Moved from https://github.com/ONSdigital/go-launch-a-survey/pull/136

### DIFF
--- a/launch.go
+++ b/launch.go
@@ -148,6 +148,7 @@ func quickLauncherHandler(w http.ResponseWriter, r *http.Request) {
 	urlValues.Add("ru_ref", authentication.GetDefaultValues()["ru_ref"])
 	urlValues.Add("collection_exercise_sid", uuid.NewV4().String())
 	urlValues.Add("case_id", uuid.NewV4().String())
+	urlValues.Add("questionnaire_id", randomNumericString(16))
 	urlValues.Add("response_id", randomNumericString(16))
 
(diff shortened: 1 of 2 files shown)


bitdivision:
This looks good to me. I can't approve it though.

launch.go
```diff
@@ -147,7 +147,7 @@ func quickLauncherHandler(w http.ResponseWriter, r *http.Request) {
 
 	urlValues.Add("ru_ref", authentication.GetDefaultValues()["ru_ref"])
 	urlValues.Add("collection_exercise_sid", uuid.NewV4().String())
-	urlValues.Add("case_id", uuid.NewV4().String())
```
bitdivision: From @MebinAbraham :

Do we want to keep this anyway since it is optional now in runner.

bitdivision: I think the answer is probably yes, since there is still runner code that uses case_id, and I don't think it fits as questionnaire metadata. May be useful to test with runner.

iwootten: I've added this back in for now. There should be unit tests within runner that cover submitter behaviour. I'm guessing once case_id is no longer part of the payload it can be removed entirely.

PR #10, opened 2019-06-26
Allow launching census with survey / case_type / region_code

bitdivision:
In combination with [runner PR](https://github.com/ONSdigital/eq-survey-runner/pull/2139).

Also uses schema_name instead of eq_id / form_type

Can be used to test changes being made in runner
around the census specific claims.

For any schemas starting with census / ccs, launcher will
convert the schema name into a survey / case_type / region_code

The UUID package being used was not maintained, and had a minor
version change with a breaking API change. I have switched to a
fork which is being maintained.

language_code now defaults to en on quick launch

### How to review
- Check that when a census schema is launched, the `survey` & `case_type` & `region_code` appear in the claims. Ensure schema_name does not.
- When launching a test schema, the schema_name should appear in the claims. The census claims should not.
- Check that quick-launch works with a schema having a file extension and no file extension
- Check that quick-launch works with a `schema_name` specified in the schema itself and a different filename.
- Ensure language_code defaults to 'en' on quick-launch.
- Make sure there are no references to eq_id and form_type left.

### DIFF
--- a/authentication/auth.go
+++ b/authentication/auth.go
@@ -14,7 +14,7 @@ import (
 	"github.com/ONSdigital/eq-questionnaire-launcher/clients"
 	"github.com/ONSdigital/eq-questionnaire-launcher/settings"
 	"github.com/ONSdigital/eq-questionnaire-launcher/surveys"
-	"github.com/satori/go.uuid"
+	"github.com/gofrs/uuid"
 	"gopkg.in/square/go-jose.v2"
 	"gopkg.in/square/go-jose.v2/json"
 	"gopkg.in/square/go-jose.v2/jwt"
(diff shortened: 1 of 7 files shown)


authentication/auth.go
```diff
@@ -362,14 +385,39 @@ func GenerateTokenFromDefaults(surveyURL string, accountServiceURL string, accou
...
+// This function can be removed after census claims are removed.
+func TransformSchemaParamsToName(postValues url.Values) string {
+	if postValues.Get("schema_name") != "" {
+		return postValues["schema_name"][0]
+	}
+
+	caseTypeMap := map[string]string{
+		"HH": "household",
+		"HI": "individual",
+		"CE": "communical_establishment",
```
bambrp: communical -> communal?

bitdivision:
There seems to be a problem with quick launching via a "schema_name" field in the schema at the moment. I'm working on that now.

launch.go
```diff
@@ -143,13 +145,17 @@ func quickLauncherHandler(w http.ResponseWriter, r *http.Request) {
...
 	surveyURL := urlValues.Get("url")
+	defaultValues := authentication.GetDefaultValues()
 	log.Println("Quick launch request received", surveyURL)
 
-	urlValues.Add("ru_ref", authentication.GetDefaultValues()["ru_ref"])
-	urlValues.Add("collection_exercise_sid", uuid.NewV4().String())
-	urlValues.Add("case_id", uuid.NewV4().String())
+	urlValues.Add("ru_ref", defaultValues["ru_ref"])
+	collectionExerciseSid, _ := uuid.NewV4()
+	caseID, _ := uuid.NewV4()
```
iwootten: Seems to be a mix of cases on vars with 'ID'?

bitdivision: Yeah, I think this might be an exception specific to go? My editor auto linted it from `caseId` to `caseID` (using golint). In the case above, it is an `Sid` which doesn't have an exception I guess.

bitdivision:
Should be merged with the runner PR.

PR #12, opened 2019-07-16
Add channel field to launcher metadata

bambrp:
To test the additional claim that will be added by RH and required downstream, channel has been added to the launcher page. Valid values are FIELD, RH, CC or AD only.

### To test
In order to test this, you will need to pull the _eq-2754-add-channel-to-submission_ branch of eq-survey-runner, which has the necessary changes to validate and submit the new claim downstream.

You can use the `/debug/submission` API endpoint to examine the JSON of what would be sent downstream.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,4 +27,4 @@ _testmain.go
 vendor/
 
 # binaries
-go-launch-a-survey
+eq-questionnaire-launcher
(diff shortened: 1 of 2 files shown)


bambrp:
[Add channel, case_type & region_code to metadata for downstream (S)](https://trello.com/c/VeQ2MjkS/2754-add-channel-casetype-regioncode-to-metadata-for-downstream-s)

PR #15, opened 2019-10-07
CCS schema branch

petechd:
CCS schemas had to be removed from Business to its own branch in launcher. Moving it to CCS branch is not best possible solution(it is in fact Census schema) but the way schemas are created in runner's "make build schemas" script, "census_ccs" or "census_cs" combinations are not allowed. 
### How to review
Make sure that port 8000 is not being used. Install Go as described in README file and run launcher using it, start runner and supporting services, then check if you can start CCS schemas from launcher's menu and it's on a correct branch.

### DIFF
--- a/surveys/surveys.go
+++ b/surveys/surveys.go
@@ -183,17 +183,17 @@ func FindSurveyByName(name string) LauncherSchema {
 			return survey
 		}
 	}
+	for _, survey := range availableSchemas.CCS {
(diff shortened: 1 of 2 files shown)


JamesGardiner:
[Add CCS to go launcher](https://trello.com/c/7ZyQyhr3/3436-add-ccs-to-go-launcher)

PR #22, opened 2020-10-09
Handle index error when accessing schema name

MebinAbraham:
Avoids index error when schema name does not exist in the claims.

### How to review 
Are you able to successfully quick launch a census schema?

If you use docker, update the image tag to: 
```        
image: "onsdigital/eq-questionnaire-launcher:handle-index-error-when-check-schema-name"
```


ajmaddaford (approved):
Tested against the docker image 👍

LJBabbage (approved):
Works for me

PR #26, opened 2021-04-19
Migrate from Travis to Github actions

ajmaddaford:
Migrates the Dockerhub image publishing from Travis to Github Actions.

### How to review 
Check that the workflows are correct.


MebinAbraham (approved):
Looks good, docker image as expected.

The docker push command could be removed from the PR workflow file and the master one could be a generic push set up that pushed to docker hub on both merges to master and on PRs. It means you need an extra command to say "latest" if master etc but would avoid the duplication. However, approved as it is consistent with our other repos.

JamesGardiner (approved):

petechd (approved):

PR #35, opened 2021-12-02
Parse launcher schema by survey type

MebinAbraham:
Updates launcher to work with https://github.com/ONSdigital/eq-questionnaire-runner/pull/731
Runner schemas are now grouped by survey type. The changes have been kept to a minimum, could be refactored further but didn't think it was worth it.

### How to review 
- Run this branch of launcher locally or point your launcher to this branch in runner's docker compose file.
- Ensure launcher works as expected and no existing functionality have been broken..
- Add a new survey type dir into runner `schemas/` dir with an appropriate schema and ensure launcher is able to pick it up.

### DIFF
--- a/surveys/surveys.go
+++ b/surveys/surveys.go
@@ -17,16 +17,9 @@ import (
 
 // LauncherSchema is a representation of a schema in the Launcher
 type LauncherSchema struct {
-	Name string
-	URL  string
-}
-
-// LauncherSchemas is a separation of Test and Live schemas
-type LauncherSchemas struct {
-	Business []LauncherSchema
-	Social   []LauncherSchema
-	Test     []LauncherSchema
-	Other    []LauncherSchema
+	Name       string
+	SurveyType string
+	URL        string
 }
 
 // RegisterResponse is the response from the eq-survey-register request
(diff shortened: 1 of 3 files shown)


templates/launch.html
```diff
@@ -157,11 +145,12 @@ <h3>Business Survey Metadata</h3>
         document.getElementById("flush-btn").disabled = true;
 
         const schema_name = document.getElementById("schema_name").value
+        const survey_type = document.querySelector(`#schema_name option[value=${schema_name}]`).dataset.surveyType
```
petechd, line 148: Can we be consequent here and rewrite the document so we use `querySelector` everywhere or use older element selection method in the whole template?

rmccar: In this instance `querySelector` was needed because of what we are trying to select. I don't personally have a problem using both as I think both are still current and valid ways of selecting elements and we are using them for different purposes in this document. But Im happy to change it and just use `querySelector`

petechd: `querySelector` is more modern API compared to `getElementById` and is explicit, you use css id and class selectors. I would keep it updated and consistent.

rmccar: Yeah it is also more explicit too to be fair. I've changed it now but can remove the commit?

MebinAbraham: I think old api was used, so launcher can run in old browsers. I think querySelector has fairly large support in most browsers, so should be fine.

> I've changed it now but can remove the commit?

No, it is perfectly fine to keep.

> I would keep it updated and consistent.

Agreed, but there is a lot that could be done to the JS here but not strictly within scope of this. Happy to keep as is if it hasn't regressed anything.

rmccar: Yeah, I genuinely don't think it makes a difference, `querySelector` is a bit newer but `getElementById` is still valid, current and used. It's not obsolete or deprecated, it's just a slightly different way of doing it. As for using one API or two I don't think that makes a difference either to me in this case. But I don't have strong enough feelings on it either way so I made the change

petechd: I just followed the `JavaScript: The Definitive Guide, 7th Edition`, these are author's words about `getElementById`👍

petechd: But @MebinAbraham's comment about old browsers support is valid, `querySelector` is less compatible than the old method.

rmccar: Oh ok, I had a bit of a google to check and couldn't find anything current saying not to use it. Doesn't matter now anyway

petechd: Some old browsers lack compatibility (IE and Firefox up to 2007/2008):
Screenshot 2021-12-07 at 09 13 07

rmccar: Yeah that was one reason I found to go with `getElementById` instead but as this is a dev tool that shouldn't really matter

petechd (approved):

LJBabbage (approved):
So I tested social and business, added another survey type and added metadata, all was good. I even added a welsh version into one of the business surveys and it worked as before. I did see a problem with json file name which I will put on the other PR

surveys/surveys.go
```diff
@@ -170,25 +161,15 @@ func getAvailableSchemasFromRunner() []LauncherSchema {
...
-			return survey
-		}
-	}
-	for _, survey := range availableSchemas.Other {
-		if survey.Name == name {
-			return survey
+	for _, schemasBySurveyType := range availableSchemas {
+		for _, schema := range schemasBySurveyType {
+			log.Print(schema.Name)
+			log.Print(name)
```
MebinAbraham, line 167: Can be removed.
```suggestion
```

PR #40, opened 2022-04-20
Ability to specify schema URL

MebinAbraham:
- Added the ability to specify a schema URL instead of selecting a schema from runner.
- Specifying a URL means that the runner claims will include a `survey_url` field which runner will use to launch.

> Note: Runner current implementation requires `schema_name` or `eq_id` and `form_type`. Therefore, when launching using a URL, schema name is passed for test schemas and for non test schema eq_id and form_type are passed. This is likely to change in future iterations.

### How to review 

Run this branch via docker or locally.
***Note: You need to clear your browsers cache since the CSS has changed.***

Ensure:
- you are able to specify a schema URL and launch the questionnaire. Validate runner did in fact launch using `survey_url` and was not loaded from file system. Use a schema that doesn't exist in runner.
- the existing behaviour of selecting from the schema list is unaffected.
- when launching using a schema URL, runner's downstream payload contains `collection.schema_name`.

### DIFF
--- a/templates/launch.html
+++ b/templates/launch.html
@@ -25,6 +25,7 @@
     <p>----------</p>
 
     <strong><u>Launch by Schema URL</u></strong>
+
     <div class="field-container">
         <span>
             <label for="schema-url-survey-type">Survey Type</label>
@@ -41,7 +42,6 @@
         <span>
             <label for="schema-url">Schema URL</label>
             <input id="schema-url" name="survey_url" type="text" class="qa-schema_url">
-            <input type="submit" onClick="return loadMetadataForSchemaUrl()" value="Load Schema" class="qa-btn-submit-dev btn" id="schema-url-btn"/>
         </span>
     </div>
 
@@ -50,6 +50,10 @@
             <label for="schema-url-name-input">Schema Name</label>
             <input id="schema-url-name-input" name="schema_name" type="text" class="qa-schema_url_name" disabled>
         </span>
+    </div>
+
+    <div class="field-container">
+        <input type="submit" onClick="return loadMetadataForSchemaUrl()" value="Load Schema" class="qa-btn-submit-dev btn" id="schema-url-btn"/>
     </div>
 
     <p>----------</p>
--- a/templates/launch.html
+++ b/templates/launch.html
@@ -178,6 +178,8 @@
     }
 
     function clearAndDisableSchemaUrlSelection() {
+        document.querySelector("#schema-url-btn").disabled = true;
+
         let schemaUrlElement = document.querySelector("#schema-url")
         let surveyTypeElement = document.querySelector("#schema-url-survey-type")
 
@@ -205,11 +207,6 @@
     }
 
     function loadMetadataForSchemaUrl() {
-        if (document.querySelector("#schema_name").selectedIndex !== 0) {
-            document.querySelector("#schema-url-btn").disabled = true;
-            return false
-        }
-
         let schemaUrl = document.querySelector("#schema-url").value
         if (!schemaUrl || !schemaUrl.endsWith(".json")) {
             alert("Schema URL not provided or is not valid.\n" +
--- a/templates/launch.html
+++ b/templates/launch.html
@@ -214,7 +214,7 @@
         }
 
         let surveyType = document.querySelector("#schema-url-survey-type").value
-        if (surveyType === "Select Survey Type") {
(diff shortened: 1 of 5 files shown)


templates/launch.html
```diff
@@ -21,12 +22,44 @@ <h1>Launch a survey</h1>
...
+
+    <div class="field-container" id="schema-url-name">
+        <span>
+            <label for="schema-url-name-input">Schema Name</label>
+            <input id="schema-url-name-input" name="schema_name" type="text" class="qa-schema_url_name" disabled>
+        </span>
+    </div>
+
+    <p>----------</p>
+
```
petechd, line 56:
```suggestion
    <strong><u>Launch by Schema URL</u></strong>
    <div class="field-container">
        <span>
            <label for="schema-url-survey-type">Survey Type</label>
            <select name="survey-types" id="schema-url-survey-type">
                <option selected disabled>Select Survey Type</option>
                {{ range $surveyType, $schemasList := .Schemas }}
                <option name="{{ $surveyType }}" value="{{ $surveyType }}" data-schema-url-survey-type="{{ $surveyType }}">{{ $surveyType }}</option>
            {{ end }}
            </select>
        </span>
    </div>

    <div class="field-container">
        <span>
            <label for="schema-url">Schema URL</label>
            <input id="schema-url" name="survey_url" type="text" class="qa-schema_url">
        </span>
    </div>

    <div class="field-container" id="schema-url-name">
        <span>
            <label for="schema-url-name-input">Schema Name</label>
            <input id="schema-url-name-input" name="schema_name" type="text" class="qa-schema_url_name" disabled>
        </span>
    </div>
    <div class="field-container">
        <input type="submit" onClick="return loadMetadataForSchemaUrl()" value="Load Schema" class="qa-btn-submit-dev btn" id="schema-url-btn"/>
    </div>
    <p>----------</p>

```

So it's similar to the existing "design".

templates/launch.html
```diff
@@ -6,9 +6,10 @@ <h1>Launch a survey</h1>
 
 <form action="" method="POST" xmlns="http://www.w3.org/1999/html" onsubmit="validateForm()">
 
+    <strong><u>Launch by Schema Name</u></strong>
```
petechd, line 9: Unarticulated annotation is used for styling text differently, misspelled words etc. Can be confused with a link, but up to you.

templates/launch.html
```diff
@@ -21,12 +22,44 @@ <h1>Launch a survey</h1>
...
+            {{ end }}
+            </select>
+        </span>
+    </div>
+
+    <div class="field-container">
+        <span>
+            <label for="schema-url">Schema URL</label>
+            <input id="schema-url" name="survey_url" type="text" class="qa-schema_url">
+            <input type="submit" onClick="return loadMetadataForSchemaUrl()" value="Load Schema" class="qa-btn-submit-dev btn" id="schema-url-btn"/>
```
petechd, line 44: Can we add similar window alert (onclick event) to other submit buttons on this page since we are making changes here? Currently when no schema is selected and we press submit button we are not getting prompted.

templates/launch.html
```diff
@@ -140,19 +173,81 @@ <h3>${survey_type} Survey Metadata</h3>
...
 
-    function loadMetadata() {
+    function clearAndDisableSchemaUrlSelection() {
+        let schemaUrlElement = document.querySelector("#schema-url")
+        let surveyTypeElement = document.querySelector("#schema-url-survey-type")
+
+        surveyTypeElement.selectedIndex = 0
+        schemaUrlElement.value = ""
+        surveyTypeElement.disabled = true
+        schemaUrlElement.disabled = true
```
petechd, line 183: Can we disable the "Load schema" button here as well?

templates/launch.html
MebinAbraham: Updated.

templates/launch.html
MebinAbraham: I've updated the CSS, the buttons are disabled to start with, so the colour has been updated. You will need to clear cache to see the change.

templates/launch.html
MebinAbraham: Was being done but in the wrong place, have updated.

templates/launch.html
MebinAbraham: Should be fine for a dev frontend that will soon get a facelift anyway.

petechd (approved):

berroar (approved):

PR #41, opened 2022-07-14
Add v2 launch button

berroar:
Describe what you have changed and why, link to other PRs or Issues as appropriate.

### How to review 
Describe the steps required to test the changes (include screenshots if appropriate).


berroar:
[Implement generic launch pattern in Launcher: v2 (3 SP)](https://trello.com/c/l8sSFPIf/5362-implement-generic-launch-pattern-in-launcher-v2-3-sp)

rmccar (approved):

launch.go
```diff
@@ -121,20 +121,25 @@ func getAccountServiceURL(r *http.Request) string {
 func redirectURL(w http.ResponseWriter, r *http.Request) {
 	hostURL := settings.Get("SURVEY_RUNNER_URL")
 
-	token, err := authentication.GenerateTokenFromPost(r.PostForm)
+	launchActionv1 := r.PostForm.Get("action_launch_v1")
+    launchActionv2 := r.PostForm.Get("action_launch_v2")
+
+	tokenV1, err := authentication.GenerateTokenFromPost(r.PostForm, false)
+	tokenV2, err := authentication.GenerateTokenFromPost(r.PostForm, true)
```
MebinAbraham, line 128: How come this is called twice with 2 vars?. We only need to generate one or the other.

authentication/auth.go
```diff
@@ -402,15 +406,15 @@ func TransformSchemaParamsToName(postValues url.Values) string {
 }
 
 // GenerateTokenFromPost converts a set of POST values into a JWT
-func GenerateTokenFromPost(postValues url.Values) (string, string) {
+func GenerateTokenFromPost(postValues url.Values, launchV2 bool) (string, string) {
```
MebinAbraham, line 409: - The actual token structure need updating to conform to the v2 data format as documented. All this is doing is sending the `version` key in the URL which is just 1 of the difference. 
- The actual structure of the payload is different between v1/v2 which launcher needs to handle based on the version we want to launch. Consult the docs for the differences and the struct itself. For example v1 would be a flat key value pair, whereas for v2, the metadata is nested.

MebinAbraham (changes requested):
Needs some work. The actual launch token structures needs to be handled.

launch.go
berroar: Updated now.

authentication/auth.go
berroar: Updated the structures for v1/v2 and business/ad-hoc.

templates/launch.html
```diff
@@ -177,6 +189,19 @@ <h3>${survey_type} Survey Metadata</h3>
...
+        let schemaNameElement = document.querySelector("#schema_name")
+        schemaNameElement.selectedIndex = 0
+
+        let launchPattern = document.querySelector("#launch_pattern").value
+
+        if (launchPattern === "v1") {
+            document.querySelectorAll(".Social").forEach(function(element) {element.setAttribute("disabled", "disabled")})
+        } else {
+            document.querySelectorAll(".Social").forEach(function(element) {element.removeAttribute("disabled")})
+        }
```
petechd, line 202:
```suggestion
        }
        
        const schemaName = schemaNameElement.value

        if (schemaName !== "Select Schema" && launchPattern === "v1") {
            document.querySelector("#submit-btn-v1").disabled = false;
            document.querySelector("#submit-btn-v2").disabled = true;
        }
        else if (schemaName !== "Select Schema" && launchPattern === "v2") {
            document.querySelector("#submit-btn-v2").disabled = false;
            document.querySelector("#submit-btn-v1").disabled = true;
        }
        else {
            document.querySelector("#submit-btn-v2").disabled = true;
            document.querySelector("#submit-btn-v1").disabled = true;
        }
```

As we discussed earlier, both buttons must disappear when no schema is selected after you change the launch pattern version.

MebinAbraham (commented):
Some findings/suggestions from a functional perspective. Not looked at code yet.
- I think we should disable or hide everything until the launch pattern is selected
- I would only display 1 button, you know which you need based on the launch pattern
- It doesn't seem to work using the load schema from URL button flow.
- Social theme is still outputting non-social claims for v2. i.e social theme schema on v2 shouldn't be including things like `ru_ref`, it should match the schema definition.
- `eq_id` shouldn't be displayed for v2 as that is only used for v1. 
- The `Survey Metadata` section is showing empty fields for social schemas. Only relevant fields should be shown.
- Sometimes the `survey_metadata.data` is empty. i.e 
```js
 'schema_name': 'social_demo',
 'survey_metadata': {'data': {}, 'receipting_keys': ['questionnaire_id']},
```

berroar:
> * The `Survey Metadata` section is showing empty fields for social schemas. Only relevant fields should be shown.
> * Sometimes the `survey_metadata.data` is empty. i.e

This is being confused by the fact that the `social_demo` schema we have is incorrect, it's theme is `default` and the survey claims are all `business` specific, which is why they are not being output.

authentication/auth.go
```diff
@@ -158,6 +207,77 @@ func generateClaims(claimValues map[string][]string, launcherSchema surveys.Laun
...
+						}
+					}
+				} else {
+					claims[key] = value
+				}
+			}
+		}
+		surveyMetadata["data"] = data
+		claims["survey_metadata"] = surveyMetadata
+	}
```
MebinAbraham, line 269: This is quite difficult to digest due to the indentation. I would break this up. It would be nice to simple have a `getSurveyMetadataFromClaims` function.

- Why do we need both `isSurveyMetadataSocial` and `isSurveyMetadataBusiness`?. We just need to know if it is survey metadata?. The UI shouldn't allow posting business claims if the schema was social. Only reason we would need to know if it is social vs business is if the same key had different uses?. Let me know if I misunderstood.

authentication/auth.go
```diff
@@ -219,11 +339,18 @@ func launcherSchemaFromURL(url string) (launcherSchema surveys.LauncherSchema, e
...
 	}
 
-	log.Println("Quicklaunch schema_name set to: ", schemaName)
+	surveyType := ""
+
+	if schema.SurveyType == "social" {
+		surveyType = "social"
+	} else {
+		surveyType = "business"
+	}
```
MebinAbraham, line 348: This wouldn't work for anything else. At this point you already have access to the schema JSON itself, so this should be set based on `theme` inside the schema.

authentication/auth.go
```diff
@@ -388,6 +515,72 @@ func GenerateTokenFromDefaults(schemaURL string, accountServiceURL string, accou
...
+// GenerateTokenFromDefaultsV2 coverts a set of DEFAULT values into a JWT
+func GenerateTokenFromDefaultsV2(schemaURL string, accountServiceURL string, accountServiceLogOutURL string, urlValues url.Values) (token string, error string) {
+	launcherSchema, validationError := launcherSchemaFromURL(schemaURL)
+	if validationError != "" {
+		return "", validationError
+	}
+
+	claims := make(map[string]interface{})
+	urlValues["account_service_url"] = []string{accountServiceURL}
+	urlValues["account_service_log_out_url"] = []string{accountServiceLogOutURL}
```
MebinAbraham, line 527: This doesn't exist in v2.

authentication/auth.go
```diff
@@ -388,6 +515,72 @@ func GenerateTokenFromDefaults(schemaURL string, accountServiceURL string, accou
...
+	}
+
+	initialData := surveyMetadata["data"].(map[string]interface{})
+
+	for key, value := range initialData {
+		updatedData[key] = value
+	}
+
+	surveyMetadata["data"] = updatedData
+	claims["survey_metadata"] = surveyMetadata
```
MebinAbraham, line 564: I'm a bit confused as to what's going on here?. Is this not essentially reconstructing the format you already had?.

MebinAbraham (commented):
Loading social schema from URL is still showing business fields such as ru_ref.

templates/launch.html
```diff
@@ -26,7 +37,7 @@ <h1>Launch a survey</h1>
 
     <strong><u>Launch by Schema URL</u></strong>
 
-    <div class="field-container">
+    <div class="field-container" style="display: none;">
```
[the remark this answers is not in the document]
berroar: Added classes and made everything but launch pattern and schema selection visible.

authentication/auth.go
berroar: I've refactored this now, I was able to removed both methods now I updated the js to stop ru_ref from being passed.

authentication/auth.go
berroar: This is because the quicklauncher code follows a different path and loads the token from the default values

authentication/auth.go
berroar: removed this now

authentication/auth.go
berroar: I removed this now and am just using schema.SurveyType which comes from the schema.

templates/launch.html
```diff
@@ -160,21 +172,71 @@ <h3>Runner Data</h3>
         document.querySelector('#survey_metadata_fields').innerHTML = ""
     }
 
+    function showAllDivs() {
+        document.querySelector("#launch_pattern").setAttribute("disabled", "disabled")
+
+        document.querySelectorAll("#survey_metadata, .hidden-field-container, .group-title", ).forEach(function (el) {
+            el.style.display = 'block';
+        });
```
petechd, line 180: I would use same convention throughout the file so for every `forEach` use `el` or `element` but not both at the same time:
```suggestion
        document.querySelectorAll("#survey_metadata, .hidden-field-container, .group-title", ).forEach(function (element) {
            element.style.display = 'block';
        });
```

authentication/auth.go
```diff
@@ -119,6 +120,66 @@ type Metadata struct {
...
+		"schema_name",
+		"account_service_url":
+		return true
+	}
+	return false
+}
+
+func getSurveyMetadataFromClaims(claimValues map[string][]string, data map[string]interface{}, claims map[string]interface{}, surveyMetadata map[string]interface{}) {
+	for key, value := range claimValues {
+		if isSurveyMetadata(key) == true {
```
petechd, line 163: Here you can make it shorter as you did on line 167
```suggestion
		if isSurveyMetadata(key) {
```

authentication/auth.go
```diff
@@ -119,6 +120,66 @@ type Metadata struct {
...
+				if isRequiredMetadata(key) {
+					if value[0] != "" {
+						claims[key] = value[0]
+					}
+				}
+			}
+			if key == "roles" {
+				claims[key] = value
+			}
+		}
```
petechd, line 176: Shortened syntax:
```suggestion
		} else if key != "roles" && isRequiredMetadata(key) && value[0] != "" {
			claims[key] = value[0]} else if key == "roles" {
			claims[key] = value
		}
	}
```

templates/launch.html
```diff
@@ -303,8 +374,19 @@ <h3>${survey_type} Survey Metadata</h3>
...
+                        document.querySelector("#submit-btn-v1").style.display = 'inline-block';
+                    } else {
+                        document.querySelector("#submit-btn-v1").style.display = 'none';
+                    }
+                    if (launchPattern === "v2") {
+                        document.querySelector("#submit-btn-v2").style.display = 'inline-block';
+                    } else {
+                        document.querySelector("#submit-btn-v2").style.display = 'none';
+                    }
+                    document.querySelector("#flush-btn").style.display = 'inline-block';
```
petechd, line 389: Cleaner if we make it one `if-else if-else`?:
```suggestion
                    if (launchPattern === "v1") {
                        document.querySelector("#submit-btn-v1").style.display = 'inline-block';
                        document.querySelector("#submit-btn-v2").style.display = 'none';
                    }
                    else if (launchPattern === "v2") {
                        document.querySelector("#submit-btn-v1").style.display = 'none';
                        document.querySelector("#submit-btn-v2").style.display = 'inline-block';
                    }
                    else {
                        document.querySelector("#submit-btn-v2").style.display = 'none';
                        document.querySelector("#submit-btn-v2").style.display = 'none';
                    }
                    document.querySelector("#flush-btn").style.display = 'inline-block';
```

petechd:
We should refactor launcher's Javascript as it uses some antiquated methods like `XMLHttpRequest` and we use a confusing mix of variables assignments (`let`, `var`, `const`) but it's not the topic of this card.

templates/launch.html
```diff
@@ -38,69 +49,69 @@ <h1>Launch a survey</h1>
...
         </span>
     </div>
 
-    <div class="field-container" id="schema-url-name">
+    <div class="hidden-field-container" id="schema-url-name">
         <span>
             <label for="schema-url-name-input">Schema Name</label>
             <input id="schema-url-name-input" name="schema_name" type="text" class="qa-schema_url_name" disabled>
         </span>
     </div>
```
MebinAbraham, line 64: Can this now not be removed?. In v2 spec, if you have a URL, you don't need `schema_name`. `schema_name / eq_id/form_type` was mandatory in v1.

templates/launch.html
MebinAbraham: I think most of this should be removed. I would just create a function to toggle the button after loading from schema name or schema url. i.e


```js
    displayLaunchButton = () => {
        let launchPattern = document.querySelector("#launch_pattern").value
        document.querySelector(`#submit-btn-${launchPattern}`).style.display = 'inline-block';
        document.querySelector("#flush-btn").style.display = 'inline-block';
    }
```

Then just call this at the end of `loadMetadataForSchemaName()` and `loadMetadataForSchemaUrl`

templates/launch.html
```diff
@@ -160,21 +172,71 @@ <h3>Runner Data</h3>
...
+                <div class="field-container" class="visible-block">
+                    <label for="form_type">form_type</label>
+                    <input id="form_type" name="form_type" type="text" value="${formTypeValue}" class="qa-form_type">
+                </div>
+            `
+        }
+
+    }
+
+    function clearSchemaNameSelection() {
```
MebinAbraham, line 214: Is this function needed anymore if we are making asking you to select the launch pattern upfront, which you can't change without a page reload anyway?.

templates/launch.html
```diff
@@ -38,69 +49,69 @@ <h1>Launch a survey</h1>
...
+    <h3 class="group-title">Survey Metadata</h3>
+    <div id="survey_metadata" class="hidden-block">
         <p>--- Metadata fields will be loaded when you select or load a schema ---</p>
     </div>
 
-    <h3>Required Data</h3>
+    <h3 class="group-title">Required Data</h3>
 
-    <div class="field-container">
+    <div class="hidden-field-container" id="ru_ref_container">
```
MebinAbraham, line 82: In the v2 scenario, this is now duplicated in survey metadata bit.

I'm in favour of removing these fields, we don't need anything here that aren't top level. The only reason we may have wanted to manually set ru_ref etc is because it was used in the response_id generation which we no longer need as response_id is a required field.

MebinAbraham (commented):
- Launcher should dynamically load any metadata defined in the schema, it shouldn't be a hardcoded mapping, as it is likely to be extended at any time. i.e currently launching the social schema is launching it like a business schema.
- We need to think about loading the schema theme before launch, as currently we have no way of testing a social test schema. Previously there was no need, as the structure was the same between the two, but not any more.

templates/launch.html
```diff
@@ -160,21 +172,71 @@ <h3>Runner Data</h3>
...
+            document.querySelector("#submit-btn-v2").style.display = 'none';
+        }
+        else if (schemaName !== "Select Schema" && launchPattern === "v2") {
+            document.querySelector("#submit-btn-v2").style.display = 'inline-block';
+            document.querySelector("#submit-btn-v1").style.display = 'none';
+        }
+        else {
+            document.querySelector("#submit-btn-v2").style.display = 'none';
+            document.querySelector("#submit-btn-v1").style.display = 'none';
+        }
```
rmccar, line 239: To me something like this is more readable and tidier
```suggestion
        if (schemaName !== "Select Schema") {
            if (launchPattern === "v1") {
                document.querySelector("#submit-btn-v1").style.display = 'inline-block';
                document.querySelector("#submit-btn-v2").style.display = 'none';
            } else if (launchPattern === "v2") {
                document.querySelector("#submit-btn-v2").style.display = 'inline-block';
                document.querySelector("#submit-btn-v1").style.display = 'none';
            }
        } else {
            document.querySelectorAll("#submit-btn-v1", "#submit-btn-v2").style.display = 'none';
        }
```

templates/launch.html
```diff
@@ -238,10 +308,11 @@ <h3>${survey_type} Survey Metadata</h3>
     }
 
     function loadSurveyMetadata(schema_name, survey_type) {
-        document.querySelector("#submit-btn").disabled = true;
-        document.querySelector("#flush-btn").disabled = true;
+        document.querySelector("#submit-btn-v1").style.display = 'none';
+        document.querySelector("#submit-btn-v2").style.display = 'none';
+        document.querySelector("#flush-btn").style.display = 'none';
```
rmccar, line 313: Can reduce this to one line
```suggestion
        document.querySelectorAll("#submit-btn-v1", "#submit-btn-v2", "#flush-btn").style.display = 'none';
```

rmccar (commented):
A few, mainly minor things I noticed that we could improve on

templates/launch.html
berroar: Added a function now 👍

templates/launch.html
berroar: Removed this

templates/launch.html
berroar: Perhaps need to update the name of this, as some of this is still required, e.g. this is currently stopping social schemas from being laucnhed in v1 mode for example

templates/launch.html
berroar: Removed this now

templates/launch.html
berroar: Updated this but had to use `querySelector` rather than `querySelectorAll` in the else

static/css/main.css
```diff
@@ -16,10 +16,33 @@ label { display:inline-block; font-weight: 600; margin-bottom: 0.25rem; width:20
...
+
 .field-container span {
     position: relative;
 }
 
+.group-title{
+    display:none;
+}
+
+.visible-block{
```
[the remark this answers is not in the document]
berroar: I have updated and removed this 👍

static/css/main.css
```diff
@@ -16,10 +16,33 @@ label { display:inline-block; font-weight: 600; margin-bottom: 0.25rem; width:20
     position: relative;
 }
 
+.hidden-field-container {
```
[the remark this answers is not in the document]
berroar: Refactored these as discussed 👍

static/css/main.css
```diff
@@ -11,15 +11,27 @@ label { display:inline-block; font-weight: 600; margin-bottom: 0.25rem; width:20
...
 }
 
+.group-title{
+    display:none;
+}
+
+.metadata-fields{
+    display:none;
+}
+
```
rmccar, line 34: I think these should be renamed `.field-container--hidden`, `.group-title--hidden` and `.metadata-fields` to make it more descriptive and fit with BEM class naming. These will then need to be updated in the html. We can also make it much simpler buy setting them all at the same time like this:
```suggestion
.field-container--hidden, .metadata-fields-hidden, .group-title-hidden { 
    display:none;
}

.field-container__span {
    position: relative;
}
```

rmccar:
This class can be removed its not being used
```
#schema-url-name {
    display: none;
}
```

static/css/main.css
```diff
@@ -11,15 +11,27 @@ label { display:inline-block; font-weight: 600; margin-bottom: 0.25rem; width:20
...
 #schema-url-btn {
```
rmccar, line 35: This class should be renamed `.btn--small` and have every line bar the last 3 removed. Then this class can be added to the `schema-url-btn` in the html. It would remove a lot of duplication alongside removing the other classes I've suggested

rmccar:
The `field-wrap` class can be removed and removed along with the `<div>` from the html. It isn't doing anything

rmccar:
I also think it would be nice to have a bit of feedback on what you can and cant click on. So I would add:
- `cursor: pointer;` to the `.btn` class
- `cursor: default;` to the `.btn:disabled` class
- `cursor: pointer;` to the `select` class
- and add this new class:
```
select:disabled {
    cursor: default;
}
```

templates/launch.html
```diff
@@ -326,13 +381,13 @@ <h3>${survey_type} Survey Metadata</h3>
...
         var result = '';
         var chars = '0123456789';
         for (var i = 16; i > 0; --i) {
             result += chars[Math.round(Math.random() * (chars.length - 1))];
         }
-        document.querySelector(`#${el_id}`).value = result;
+        document.querySelector(`#response_id`).value = result;
     }
 
     function ruref(el_id) {
```
rmccar, line 393: This function is causing an error in the console because you have removed the `#ru_ref` element. So it, alongside the call for it should probably be removed

berroar: Removed now

static/css/main.css
berroar: Added now 👍

templates/launch.html
```diff
@@ -202,8 +255,17 @@ <h3>${survey_type} Survey Metadata</h3>
...
+            hideNonSocialMetadata()
+        }
+
         loadSurveyMetadata(schemaName, surveyType)
         loadSchemaMetadata(schemaName, null)
+        displayLaunchButton()
+    }
+
+    function hideNonSocialMetadata() {
+        document.querySelector("#ru_ref_container").innerHTML=""
```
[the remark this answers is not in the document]
berroar: removed now

static/css/main.css
```diff
@@ -54,6 +54,11 @@ select {
 
 select[multiple] {
     background:none;
+    cursor: pointer;
```
rmccar, line 57: This line should be moved to the class above

static/css/main.css
berroar: Updated now 👍

templates/launch.html
```diff
@@ -160,21 +156,57 @@ <h3>Runner Data</h3>
         document.querySelector('#survey_metadata_fields').innerHTML = ""
     }
 
+    function showAllDivs() {
+        document.querySelector("#launch_pattern").setAttribute("disabled", "disabled")
+
+        document.querySelectorAll("#survey_metadata, .field-container--hidden, .group-title--hidden", ).forEach(function (element) {
```
rmccar, line 162:
```suggestion
        document.querySelectorAll("#survey_metadata, .field-container--hidden, .group-title--hidden").forEach(function (element) {
```

rmccar, line 162:
```suggestion
        document.querySelectorAll("#survey_metadata, .field-container--hidden, .group-title--hidden").forEach(function (element) {
```

templates/launch.html
```diff
@@ -1,155 +1,151 @@
...
+
+        <div class="field-container field-container--hidden">
+            <label for="roles">Roles</label>
+            <select id="roles" name="roles" multiple="multiple" class="qa-roles">
+                <option name="flusher" value="flusher">flusher</option>
+                <option name="dumper" value="dumper" selected="selected">dumper</option>
+            </select>
+        </div>
+
+        <div class="field-container  field-container--hidden">
```
rmccar, line 132:
```suggestion
        <div class="field-container field-container--hidden">
```

rmccar (commented):
Couple of really minor things, other than that looks good

petechd (commented):
When I select launch pattern from dropdown but don’t pick any schema and press the button to open, launcher crashes. Button should just be hidden until schema + version selected?

berroar:
> When I select launch pattern from dropdown but I don’t pick any schema and press the button to open, launcher crashes. Button should be hidden until schema + version selected? Or a prompt like when trying to press load schema from url but no url provided?

Fixed now 👍

authentication/auth.go
```diff
@@ -461,50 +662,82 @@ func GetRequiredMetadata(launcherSchema surveys.LauncherSchema) ([]Metadata, str
...
+	return false
 }
 
 // GetDefaultValues Returns a map of default values for metadata keys
 func GetDefaultValues() map[string]string {
-
 	defaults := make(map[string]string)
-
 	collectionExerciseSid, _ := uuid.NewV4()
+	questionnaireId, _ := uuid.NewV4()
```
MebinAbraham, line 735: Questionnaire ID needs to be a 16 digit string. The validation for this is correct in the schemas repo. The md doc uses UUID but as per validation, this will need to be a 16 digit string.

rmccar:
Currently you can select a schema without selecting a version and the open survey button appears allowing you to click it which crashes launcher. I think the buttons should only be shown when both have been selected

rmccar:
The Launch by Schema URL part doesn't seem to work correctly the load schema button doesn't appear when entering a url and the open schema buttons appear when you hit enter in the url box but nothing else. When You get the correct buttons to appear by using the version and schema dropdowns above the functionality does work as expected though

templates/launch.html
```diff
@@ -1,155 +1,150 @@
...
+<div>
+    <form action="" method="POST" xmlns="http://www.w3.org/1999/html" onsubmit="validateForm()">
+
+        <strong><u>Launch Pattern</u></strong>
+        <div class="field-container">
+            <label for="launch_pattern">Version</label>
+            <select id="launch_pattern" name="launch_version" class="qa-select-launch-version" onchange="showAllDivs();">
+                <option selected disabled>Select Launch Pattern</option>
+                <option class="version" name="version" id="version_v1" value="v1">v1</option>
+                <option class="version" name="version" id="version_v2" value="v2">v2</option>
```
rmccar, line 15:
```suggestion
                <option value="v1">v1</option>
                <option value="v2">v2</option>
```

templates/launch.html
```diff
@@ -1,155 +1,150 @@
...
+            <label for="schema_name">Schemas</label>
+            <select id="schema_name" name="schema_name" class="qa-select-schema" onchange="loadMetadataForSchemaName(); displayLaunchButton();">
+                <option selected disabled>Select Schema</option>
+
                 {{ range $surveyType, $schemasList := .Schemas }}
-                <option name="{{ $surveyType }}" value="{{ $surveyType }}" data-schema-url-survey-type="{{ $surveyType }}">{{ $surveyType }}</option>
-            {{ end }}
+                    <optgroup label="{{ $surveyType }} Surveys">
+                        {{ range $schema := $schemasList }}
+                            <option class="{{ $surveyType }}" name="{{ $schema.Name }}" value="{{ $schema.Name }}" data-survey-type="{{ $surveyType }}">{{ $schema.Name }}</option>
```
rmccar, line 28:
```suggestion
                            <option value="{{ $schema.Name }}" data-survey-type="{{ $surveyType }}">{{ $schema.Name }}</option>
```

templates/launch.html
```diff
@@ -1,155 +1,150 @@
...
+
+        <strong><u>Launch by Schema URL</u></strong>
+
+        <div class="field-container">
+            <span class="field-container__span">
+                <label for="schema-url-survey-type">Survey Type</label>
+                <select name="survey-types" id="schema-url-survey-type">
+                    <option selected disabled>Select Survey Type</option>
+                    {{ range $surveyType, $schemasList := .Schemas }}
+                    <option name="{{ $surveyType }}" value="{{ $surveyType }}" data-schema-url-survey-type="{{ $surveyType }}">{{ $surveyType }}</option>
```
rmccar, line 45:
```suggestion
                    <option value="{{ $surveyType }}">{{ $surveyType }}</option>
```

templates/launch.html
```diff
@@ -1,155 +1,150 @@
...
+        </div>
+
+        <div class="field-container field-container--hidden">
+            <label for="language_code">Language</label>
+            <select id="language_code" name="language_code" class="qa-language-code">
+                <option name="en" value="en">English (en)</option>
+                <option name="cy" value="cy">Cymraeg (cy)</option>
+                <option name="ga" value="ga">Gaeilge (ga)</option>
+                <option name="eo" value="eo">Ulstér Scotch (eo)</option>
+                <option name="" value="">&lt;not set&gt;</option>
```
rmccar, line 120:
```suggestion
                <option value="en">English (en)</option>
                <option value="cy">Cymraeg (cy)</option>
                <option value="ga">Gaeilge (ga)</option>
                <option value="eo">Ulstér Scotch (eo)</option>
                <option value="">&lt;not set&gt;</option>
```

templates/launch.html
```diff
@@ -1,155 +1,150 @@
...
+            </select>
+        </div>
+
+        <div class="field-container field-container--hidden">
+            <label for="roles">Roles</label>
+            <select id="roles" name="roles" multiple="multiple" class="qa-roles">
+                <option name="flusher" value="flusher">flusher</option>
+                <option name="dumper" value="dumper" selected="selected">dumper</option>
```
rmccar, line 128:
```suggestion
                <option value="flusher">flusher</option>
                <option value="dumper" selected>dumper</option>
```

MebinAbraham (commented):
Quick launch needs a looking at. For example, if I launch a v2 social schema, it still has `ru_ref`.

rmccar:
> Currently you can select a schema without selecting a version and the open survey button appears allowing you to click it which crashes launcher. I think the buttons should only be shown when both have been selected. Also the code that shows buttons and extra fields should be on if something has been selected rather than on change because if you go into a survey and then use the back button in the browser, you end up with no buttons and fields showing even though there is a version and schema selected. The fields are only then shown if you change the version and the buttons only appear if you change the schema

The back button issue I mentioned here isn't resolved. Ill add some suggestions for some code that will reset the dropdowns on page load

templates/launch.html
```diff
@@ -1,155 +1,150 @@
 {{define "title"}}Launch a Survey{{end}}
 
 {{define "body"}}
+
```
rmccar, line 4:
```suggestion

<body onbeforeunload='resetDropdowns();'>
```

rmccar (commented):
These changes plus going into the `layout.html` file and removing the 2 body tags will resolve the back button dropdown issue

templates/launch.html
```diff
@@ -363,8 +382,7 @@ <h3>${survey_type} Survey Metadata</h3>
```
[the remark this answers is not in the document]
berroar: Added now 👍

templates/launch.html
```diff
@@ -1,155 +1,150 @@
...
+            <input id="account_service_log_out_url" name="account_service_log_out_url" type="text" value="{{.AccountServiceLogOutURL}}" class="qa-account_service_log_out_url">
+        </div>
+
+        <div class="field-container">
+            <input type="submit" name="action_launch" value="Open Survey" class="qa-btn-submit-dev btn button--hidden" id="submit-btn"/>
+            <input type="submit" name="action_flush" value="Flush Survey Data" class="qa-btn-submit-dev btn button--hidden" id="flush-btn"/>
+        </div>
+
+    </form>
 </div>
```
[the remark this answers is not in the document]
berroar: Added now 👍

templates/launch.html
```diff
@@ -1,157 +1,153 @@
 {{define "title"}}Launch a Survey{{end}}
 
 {{define "body"}}
+
+<body onbeforeunload='resetDropdowns();'>
```
rmccar, line 5: Everything inside this body tag should be tabbed across

templates/layout.html
```diff
@@ -7,8 +7,6 @@
     <title>{{template "title"}}</title>
     <link rel="stylesheet" href="/static/css/main.css">
 </head>
-<body>
 {{template "body" .}}
```
rmccar, line 10: Everything inside the html tags here should be tabbed across and a newline should be added

rmccar (commented):
Other than these very minor things, looks good!

templates/layout.html
berroar: Updated 👍

authentication/auth.go
```diff
@@ -461,50 +663,81 @@ func GetRequiredMetadata(launcherSchema surveys.LauncherSchema) ([]Metadata, str
...
+}
 
-	for i, value := range schema.Metadata {
-		schema.Metadata[i].Default = defaults[value.Name]
+func getMandatatoryClaims(surveyType string, defaults map[string]string) []Metadata {
+	claims := make([]Metadata, 0)
+	if surveyType == "health" || surveyType == "social" {
+		claims = []Metadata{
+			{"qid", "false", defaults["qid"]},
+			{"case_ref", "false", defaults["case_ref"]},
```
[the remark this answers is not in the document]
berroar: Dropped now 👍

rmccar (approved):

MebinAbraham (approved):

PR #44, opened 2023-01-20
Add support for PHM metadata

petechd:
This adds support for PHM metadata, works for both standard launcher and quick launch. Barcodes are generated based on name and do not have default values. `COLLEX_OPEN_DATE` and `COLLEX_END_DATE` are string types at the moment but can be changed to date types if needed (that will probably require changes in runner).

### How to review 
Add metadata fields to PHM schema, run launcher locally and check if new metadata doesn't break runner, test quick launch.

### DIFF
--- a/authentication/auth.go
+++ b/authentication/auth.go
@@ -137,7 +137,15 @@ func isSurveyMetadata(key string) bool {
 		"ru_ref",
 		"trad_as",
 		"user_id",
-		"qid":
+		"qid",
+		"PARTICIPANT_ID",
+		"COLLEX_OPEN_DATE",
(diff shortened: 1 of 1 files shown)


petechd:
[Add support to launcher for new PHM metadata (3 SP)](https://trello.com/c/44BIXVTF/5538-add-support-to-launcher-for-new-phm-metadata-3-sp)

berroar (approved):
Was able to launch PHM with this, shame these new metadata fields are upper case where as the existing ones are lower 😅

MebinAbraham (approved):
👍

PR #48, opened 2023-05-25
Ability to launch with the sds_dataset_id

berroar:

Adds the ability to launch a questionnaire schema with `sds_dataset_id` field set as Survey Metadata. This will allow us to launch surveys with Prepop data.

### How to review

To be tested alongside this PR: https://github.com/ONSdigital/eq-questionnaire-runner/pull/1114

### DIFF
--- a/authentication/auth.go
+++ b/authentication/auth.go
@@ -720,7 +720,6 @@ func getMandatatoryClaims(surveyType string, defaults map[string]string) []Metad
 			{"ru_ref", "false", defaults["ru_ref"]},
 			{"period_id", "false", defaults["period_id"]},
 			{"user_id", "false", defaults["user_id"]},
-			{"sds_dataset_id", "false", defaults["sds_dataset_id"]},
 		}
 	}
 
@@ -751,6 +750,7 @@ func stringInSlice(a string, list []string) bool {
 func GetDefaultValues() map[string]string {
 	defaults := make(map[string]string)
 	collectionExerciseSid, _ := uuid.NewV4()
+	sdsDatasetId, _ := uuid.NewV4()
 
 	var PARTICIPANT_ID = "ABC-" + fmt.Sprintf("%011d", rand.Int63n(1e11))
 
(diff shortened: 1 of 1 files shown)


berroar:
[Integrate with SDS and validate the response (Mocked API) (3 SP)](https://trello.com/c/5eyexcnO/5626-integrate-with-sds-and-validate-the-response-mocked-api-3-sp)

authentication/auth.go
```diff
@@ -779,6 +781,7 @@ func GetDefaultValues() map[string]string {
 	defaults["PARTICIPANT_ID"] = PARTICIPANT_ID
 	defaults["FIRST_NAME"] = "John"
 	defaults["TEST_QUESTIONS"] = "F"
+	defaults["sds_dataset_id"] = "001"
```
MebinAbraham, line 784: dataset id will be a uuid4

MebinAbraham: The type is right, but they haven't defined a format. UUID4 is a string, it is a string in a specific format.

MebinAbraham (commented):
I'll raise separate card for it, we should dynamically select the SDS dataset id like schemas. I'll raise a card to integrate launcher with SDS.

authentication/auth.go
```diff
@@ -779,6 +782,8 @@ func GetDefaultValues() map[string]string {
...
```
katie-gardner, line 785: I think that either the default should be None because now every survey except the supplementary data one fails on launch unless you manually remove the default

katie-gardner (changes requested):

authentication/auth.go
berroar: I had as mandatory when it shouldn't be so updated that now. It should only be present on launcher now if sds_dataset_id is present in the schema metadata.

katie-gardner: ah awesome thats an even better solution 🎉

berroar: Was possibly failing for you as I hadn't pushed the change to make sds dataset id a uuid 😅

katie-gardner: The answer might be no - not so familiar with launcher but is it possible to make the dataset id default to the one from the schema metadata?

berroar: I think the ultimate aim is to have the dataset_id generated dynamically which we will do in this ticket: https://trello.com/c/WUVvv8xO/5740-integrate-sds-with-launcher-3-sp

katie-gardner: Ah perfect - in that case, all looking good to me here

katie-gardner (approved):

kylelawsonAND (approved):
Works nicely with partner runner branch👌

Yuyuutsu (approved):
LGTM !

authentication/auth.go
katie-gardner: now uses a uuid so resolving 👍

petechd (approved):

rmccar (approved):

PR #50, opened 2023-07-19
Make v2 default

rmccar:
This is to change the version dropdown to default to v2. Just to make things a bit easier for the user.

- Also fixes console error around a null field trying to be set to null
- Tidies up some formatting
- Refactors the show hide css js for some elements to make more sense
- Fixes a bug with the default value for `ref_p_start_date` and `ref_p_end_date` not being used

### How to review
- Check that v2 is selected now when launcher is loaded
- Check that schemas load as expected when selected from dropdown and using a url

### DIFF
--- a/templates/launch.html
+++ b/templates/launch.html
@@ -343,11 +343,15 @@
                     if (sds_metadata_response?.length) {
                         supplementaryDataSets = sds_metadata_response
                         toggleSupplementaryData("show");
+                        toggleSubmitFlushButtons("show");
                         document.querySelector("#sds_dataset_id").innerHTML = sds_metadata_response.map(dataset =>
                             `<option value="${dataset.dataset_id}">${dataset.dataset_id}</option>`)
                             .join("");
-                        loadSupplementaryDataInfo()
-                        displayLaunchButton(false)
+                        loadSupplementaryDataInfo();
+                        displayLaunchButton(false);
+                    } else if (document.querySelector("#sds_dataset_id")) {
+                        document.querySelector("#sds_dataset_id").innerHTML = "";
+                        toggleSubmitFlushButtons("hide");
                     }
                 })
                 .catch(_ => {
--- a/templates/launch.html
+++ b/templates/launch.html
@@ -60,11 +60,11 @@
 
                 <p>----------</p>
 
-                <div id="survey_metadata_fields" class="metadata-fields--hidden">
+                <div id="survey_metadata_fields">
                 </div>
 
                 <h3 class="group-title--hidden">Survey Metadata</h3>
-                <div id="survey_metadata" class="metadata-fields--hidden">
+                <div id="survey_metadata">
                     <p>--- Metadata fields will be loaded when you select a version and load a schema ---</p>
                 </div>
 
--- a/templates/launch.html
+++ b/templates/launch.html
@@ -1,477 +1,477 @@
-{{define "title"}}Launch a Survey{{end}}
+{{ define "title" }}Launch a Survey{{ end }}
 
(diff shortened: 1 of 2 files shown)


rmccar:
[Make v2 default on launcher](https://trello.com/c/zkCr3GCg/5798-make-v2-default-on-launcher)

templates/launch.html
```diff
@@ -1,477 +1,477 @@
...
+                        <input id="form_type" name="form_type" type="text" value="${formTypeValue}" class="qa-form_type">
+                    </div>
+                `
+            }
+
         }
 
-    }
+        function displayLaunchButton(launchBySchemaUrl) {
+            let version = document.querySelector("#launch_pattern").value
```
katie-gardner, line 221: this variable is now redundant

templates/launch.html
```diff
@@ -1,477 +1,477 @@
...
+
+        function showAllDivs() {
+            document.querySelectorAll("#survey_metadata, .field-container--hidden, .group-title--hidden").forEach(function (element) {
+                element.style.display = 'block';
+            });
+
+            let launchPattern = document.querySelector("#launch_pattern").value
+
+            if (launchPattern === "v1") {
+                document.querySelectorAll(".Social").forEach(function(element) {element.setAttribute("disabled", "disabled")})
```
katie-gardner, line 182: this function has been removed from the `onchange` method of changing the launcher version, so I don't know if we'll get correct behaviour when switching from v2 to v1 (if we needed to) because this isn't going to happen

but at the same time I don't know when `document.querySelectorAll(".Social")` would return anything, it doesn't for social surveys... know when its used?

katie-gardner (changes requested):
all for having a default to v2 👍 
Do we have any agreed formatting rules for this repo? 😆

templates/launch.html
rmccar: I agree I cant see where that would be being used. There is no elements that I can see that have a `.social` class

templates/launch.html
```diff
@@ -373,14 +373,15 @@ <h3>Runner Data</h3>
...
                             const fieldName = metadataField["name"]
-                            let defaultValue = metadataField['default'];
 
                             if (metadataField['type'] === "date") {
                                 const currentDate = new Date()
                                 // format as yyyy-MM-dd
                                 defaultValue = currentDate.toLocaleDateString().split("/").reverse().join("-")
                             }
 
+                            defaultValue = metadataField['default'];
```
katie-gardner, line 383: ah I actually deleted this line in my PR with Kyle as I thought it was an accident, if we don't want default date I think the
```
if (metadataField['type'] === "date") {
   const currentDate = new Date()
   // format as yyyy-MM-dd
   defaultValue = currentDate.toLocaleDateString().split("/").reverse().join("-")
}
```
 can be removed to avoid confusion

katie-gardner (approved):
Two tiny comments but other than that, looks good 👍

katie-gardner (approved):
Re-approved following the new changes 👍

MebinAbraham (commented):
The formatting changes make this hard to review. It should have been separate. Plus formatting rules aren't enforced at the moment, so someelse's IDE might be formatting differently.

rmccar:
> The formatting changes make this hard to review. It should have been separate. Plus formatting rules aren't enforced at the moment, so some else's IDE might be formatting differently. Happy for the formatting changes but it would have been nice if it was in a different PR.

Yeah you're right. I'm happy to split them out into another PR. Ill do that 👍

Yuyuutsu (approved):
LGTM !

PR #52, opened 2023-07-21
Fix formatting for launch.html

rmccar:
Just fixes the code formatting for `launch.html`

### How to review
- Code looks correctly formatted
- Launcher still works as expected

### DIFF
--- a/templates/launch.html
+++ b/templates/launch.html
@@ -2,476 +2,476 @@
 
 {{define "body"}}
 
-<body onbeforeunload='resetDropdowns();'>
-    <h1>Launch a survey</h1>
-    <div>
-        <form action="" method="POST" xmlns="http://www.w3.org/1999/html" onsubmit="validateForm()">
+    <body onbeforeunload='resetDropdowns();'>
+        <h1>Launch a survey</h1>
+        <div>
+            <form action="" method="POST" xmlns="http://www.w3.org/1999/html" onsubmit="validateForm()">
 
-            <strong><u>Launch Pattern</u></strong>
-            <div class="field-container">
-                <label for="launch_pattern">Version</label>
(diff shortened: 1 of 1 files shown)


rmccar:
[Make v2 default on launcher](https://trello.com/c/zkCr3GCg/5798-make-v2-default-on-launcher)

katie-gardner:
Looks good 👍 , what formatter did you use to make the changes? I'd be in favour of this being added to the README file so others can use the same one for consistency

katie-gardner (commented):
Looks good 👍  What formatter did you use to make the changes? I'd be in favour of it being added to the README so others can use the same one for consistency

rmccar:
> Looks good 👍 What formatter did you use to make the changes? I'd be in favour of it being added to the README so others can use the same one for consistency

I didn't actually use one. We should do a spike to do some research and get one though. Like we should do the same for the jinja2 files in eQ. I just saw that it should be indented because its inside the "body" block so I made the change

MebinAbraham:
I guess it was never added since it was a minor dev repo. We can add linting and formatting using `go fmt` and `djlint` for template formating.

rmccar:
> I guess it was never added since it was a minor dev repo. We can add linting and formatting using `go fmt` and `djlint` for template formatting.
> 
> Maybe leave it as is and do it officially?

Ok ill close this PR then but add a card to the board for adding this to launcher and I think we should have one for eQ too

MebinAbraham:
> > I guess it was never added since it was a minor dev repo. We can add linting and formatting using `go fmt` and `djlint` for template formatting.
> > Maybe leave it as is and do it officially?
> 
> Ok ill close this PR then but add a card to the board for adding this to launcher and I think we should have one for eQ too

We already spiked it for eQ and there were some issues with djlint and jinja hence we didn't got forward with it. If and when they fixes the formatting issues, we can revist (it might already be). There is a spike card on the board done column, have a look

PR #55, opened 2023-08-29
Save schema and response id to local storage

petechd:
This adds js localStorage property usage to launcher. Schema name and response id are now saved and can be retrieved or deleted. New buttons added to the launcher html page and extra option of clearing local storage added, otherwise loading schema from url is permanently disabled when storage elements exist and have to be cleared manually.

### How to review
Run locally using `go run launch.go` and test all the options manually.


petechd:
[Save schema and response id to local storage in launcher (1 SP)](https://trello.com/c/2JZgQ9Jp/5803-save-schema-and-response-id-to-local-storage-in-launcher-1-sp)

templates/launch.html
```diff
@@ -460,6 +463,57 @@ <h3>Runner Data</h3>
         document.getElementById('schema-url-survey-type').selectedIndex = -1;
     }
 
+    function retrieveSchemaName() {
+        let schemaName = localStorage.getItem("schema_name");
+        if (schemaName) {
```
katie-gardner, line 468: Since its a dev tool, doesn't really matter too much, but you can avoid getting a console error when the stored schema is no longer in the list if you check the schema is in the list first
```
const availableSchemas = [...document.querySelector("#schema_name").options].map(x => x.value)
if (schemaName && availableSchemas.includes(schemaName)) {
```

katie-gardner (commented):
Working as intended - the saving is great 🎉 
Only though I have re button, is it might be more intuitive for the odd other person using launcher if it is next to the field
maybe something like 
Screenshot 2023-08-30 at 11 51 40
?
not sure what you think of something like that? but happy to go as is, either works so feel free to leave that one. If the button is at the bottom I think I'd also go with it saying "Load Previous Response ID" instead for clarity

petechd:
> Working as intended - the saving is great 🎉 Only though I have re button, is it might be more intuitive for the odd other person using launcher if it is next to the field maybe something like Screenshot 2023-08-30 at 11 51 40 ? not sure what you think of something like that? but happy to go as is, either works so feel free to leave that one. If the button is at the bottom I think I'd also go with it saying "Load Previous Response ID" instead for clarity

This probably looks better in line with that field as on your image, looks great, I will try to add it this way.

templates/launch.html
petechd: Added, thanks!

petechd:
> Working as intended - the saving is great 🎉 Only though I have re button, is it might be more intuitive for the odd other person using launcher if it is next to the field maybe something like Screenshot 2023-08-30 at 11 51 40 ? not sure what you think of something like that? but happy to go as is, either works so feel free to leave that one. If the button is at the bottom I think I'd also go with it saying "Load Previous Response ID" instead for clarity

Added now, we could potentially add some svg icon next to the existing arrow but since we were told not to overthink it we could just keep the button.

katie-gardner:
> > Working as intended - the saving is great 🎉 Only though I have re button, is it might be more intuitive for the odd other person using launcher if it is next to the field maybe something like Screenshot 2023-08-30 at 11 51 40 ? not sure what you think of something like that? but happy to go as is, either works so feel free to leave that one. If the button is at the bottom I think I'd also go with it saying "Load Previous Response ID" instead for clarity
> 
> Added now, we could potentially add some svg icon next to the existing arrow but since we were told not to overthink it we could just keep the button. The only problem now is the arrow moved and slightly to the right.

Yeah I agree 👍 arrow looks slightly wonky but not that badly - I think it could be fixed with a tiny bit of padding (probably that 0.3em) on the button and `field_container__img` and tweaking `field_container__span` but yeah probably don't need to go overboard so it does the job for me

katie-gardner (approved):
Could possibly fix CSS if its simple but extra features working nicely 👍 happy as is

templates/launch.html
```diff
@@ -460,6 +463,58 @@ <h3>Runner Data</h3>
...
+
+    function loadResponseId() {
+        document.querySelector("#response_id").value = localStorage.getItem("response_id");
+    }
+
+    function saveResponseId() {
+        localStorage.setItem("response_id", document.querySelector("#response_id").value)
+    }
+
+    function clearLocalStorage() {
```
MebinAbraham, line 505: The schema url input box is still disabled. That also needs enabling.

templates/launch.html
```diff
@@ -460,6 +463,58 @@ <h3>Runner Data</h3>
...
+        localStorage.removeItem("response_id");
+        localStorage.removeItem("schema_name");
```
MebinAbraham, line 507: We could just refresh the page after this. Then we don't have to keep adding to this.
Button could just be `Clear Local Storage and Refresh`

templates/launch.html
```diff
@@ -460,6 +463,58 @@ <h3>Runner Data</h3>
...
+        else {
+            document.querySelector("#local-storage-btn").classList.add("btn--hidden");
+        }
+    }
+
+    function retrieveResponseId() {
+        let responseId = localStorage.getItem("response_id");
+
+        if (responseId) {
+            document.querySelector("#response-id-btn").classList.remove("btn--hidden");
```
MebinAbraham, line 490: Minor, avoid the duplication where possible, multiple patterns of reusing the same thing more than once. Use a variable so the element so `let responseIdButton = document.querySelector("#response-id-btn")` etc.

PR #59, opened 2023-11-08
Add support for launching CIR Surveys

katie-gardner:
This PR adds support for launching surveys that have come from CIR. A new env var `CIR_API_BASE_URL` has been added just like the SDS one for the base url of the service. By default it is that of the mock CIR. The backend uses this to fetch all `ciMetadata` and populate a new dropdown in launcher for CIR schemas. Just like with Runner, if the service is down, launcher will continue to work, but not populate the dropdown.

On Selection of a CIR schema and survey type clicking `Load Schema` will fetch the schema from CIR and use it to populate the Survey Metadata. It will use the attributes of the option to populate CIR Metadata (version, title, description, etc.)

The existing approach to disabling fields has been removed in favour of clearing fields when switching the launch type. So that you can easily swap between CIR schemas, Schema URL and Schema Name without needing to empty local storage. Local storage saving has also been extended to save the cir schema or schema url so repeated testing via these options is easier.

### How to review
To test functionally, switch to the [Runner branch with support for CIR](https://github.com/ONSdigital/eq-questionnaire-runner/pull/1246) and the [Mock CIR branch](https://github.com/ONSdigital/eq-runner-mock-cir/pull/1) of the mock repo

Run `make load-schemas` in both runner and the Mock CIR then start launcher. You should be see all the schemas in the CIR Schemas dropdown, with their names and language. Selecting and launching one should work, and the dump and submission endpoints should show that `cir_instrument_id` is being used instead of `schema_name` or `schema_url`

Check that the launcher UI works as expected when swapping between the different types of schema launches and when refreshing the page. Additionally, check it still works for normal schemas when CIR is down.

### DIFF
--- a/static/css/main.css
+++ b/static/css/main.css
@@ -11,7 +11,7 @@ label { display:inline-block; font-weight: 600; margin-bottom: 0.25rem; width:20
     position: relative;
 }
 
-.btn--hidden, .supplementary-data--hidden {
+.btn--hidden, .supplementary-data--hidden, .cir-metadata--hidden {
     display:none;
 }
 
--- a/surveys/surveys.go
+++ b/surveys/surveys.go
@@ -143,7 +143,7 @@ func getAvailableSchemasFromRegister() []LauncherSchema {
 	return schemaList
 }
 
-func GetAvailableSchemasFromCIR() ([]CIMetadata, error) {
+func GetAvailableSchemasFromCIR() []CIMetadata {
 
 	ciMetadataList := []CIMetadata{}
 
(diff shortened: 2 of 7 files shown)


katie-gardner:
[Integrate CIR with launcher (5 SP)](https://trello.com/c/UPUPJQJK/5873-integrate-cir-with-launcher-5-sp)

static/css/main.css
```diff
@@ -11,7 +11,7 @@ label { display:inline-block; font-weight: 600; margin-bottom: 0.25rem; width:20
     position: relative;
 }
 
-.btn--hidden, .supplementary-data--hidden {
+.btn--hidden, .supplementary-data--hidden, .cir-metadata--hidden {
```
petechd, line 14: Wouldn't it make more sense to call it `.hidden` class? Instead of creating 3? Or even doing inline style: "style=display:none"?

berroar (commented):
Just been testing this PR - A bit weird that you can select survey_type "Social" and then Launch a Business Schema with no issues? Do we need the survey_type for CIR?

static/css/main.css
katie-gardner: Problem with using `.hidden` is that I then can't differentiate between them, so I can't hide just one of the dropdowns, I do need it separate. That wouldn't be an issue with inline styling but my preference is to stick with the css class as it separates the styling and logic, and I don't then need to manually do something like `display=block` or similar when undoing the hide

templates/launch.html
```diff
@@ -174,6 +246,14 @@ <h3>Runner Data</h3>
...
     }
 
+    function toggleCIRMetadata(showHide) {
+        if (showHide == "hide") {
+            document.querySelector(".cir-metadata").classList.add("cir-metadata--hidden");
+        } else if (showHide == "show") {
+            document.querySelector(".cir-metadata").classList.remove("cir-metadata--hidden");
+        }
+    }
+
```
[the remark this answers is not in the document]
katie-gardner: Happy to make that adjustment, I prefer the show type function as well 👍 done

katie-gardner:
> Just been testing this PR - A bit weird that you can select survey_type "Social" and then Launch a Business Schema with no issues? Do we need the survey_type for CIR?

I do agree with this, but I think might be one to do if the CIR has an update to include it in the metadata? Then we can really easily remove it, or infer it in a follow up

static/css/main.css
petechd: Ok, makes sense.

surveys/surveys.go
```diff
@@ -127,6 +143,39 @@ func getAvailableSchemasFromRegister() []LauncherSchema {
 	return schemaList
 }
 
+func GetAvailableSchemasFromCIR() []CIMetadata {
```
petechd, line 146: We have some inconsistency in this file now, some is camel case some not, worth changing this and `GetLauncherSchema`. Or are there some different conventions used here?

petechd (approved):

PR #62, opened 2023-12-15
Fix url schema launch

rmccar:
This makes the change to only send the schema url to runner when launching via a schema url

Also fixes the alignment of the arrow icon next to the load previous value button and tidys up the code a bit.

### How to review
- Run this launcher branch locally, see that the arrow icon is now aligned
- Launch a schema from a url and see that the schema launches. This is a url you can use to test that: `https://raw.githubusercontent.com/ONSdigital/eq-questionnaire-schemas/main/schemas/business/en/1_0005.json`
- Use the quick launch facility to launch a schema and see that the schema launches correctly without error. This is a url you can use to test that: `http://localhost:8000/quick-launch?url=https://raw.githubusercontent.com/ONSdigital/eq-questionnaire-schemas/main/schemas/business/en/1_0005.json`

### DIFF
--- a/surveys/surveys.go
+++ b/surveys/surveys.go
@@ -281,8 +281,7 @@ func GetLauncherSchema(schemaName string, schemaUrl string, cirInstrumentId stri
 	if schemaUrl != "" {
 		log.Println("Getting schema by URL: " + schemaUrl)
 		launcherSchema = LauncherSchema{
-			URL:  schemaUrl,
-			Name: schemaName,
+			URL: schemaUrl,
 		}
 	} else if cirInstrumentId != "" {
 		log.Println("Searching for schema by CIR Instrument ID: " + cirInstrumentId)
--- a/surveys/surveys.go
+++ b/surveys/surveys.go
@@ -281,7 +281,7 @@ func GetLauncherSchema(schemaName string, schemaUrl string, cirInstrumentId stri
 	if schemaUrl != "" {
 		log.Println("Getting schema by URL: " + schemaUrl)
 		launcherSchema = LauncherSchema{
-			URL: schemaUrl,
+			URL:  schemaUrl,
 			Name: schemaName,
 		}
 	} else if cirInstrumentId != "" {
--- a/surveys/surveys.go
+++ b/surveys/surveys.go
@@ -282,6 +282,7 @@ func GetLauncherSchema(schemaName string, schemaUrl string, cirInstrumentId stri
 		log.Println("Getting schema by URL: " + schemaUrl)
 		launcherSchema = LauncherSchema{
 			URL: schemaUrl,
+			Name: schemaName,
 		}
 	} else if cirInstrumentId != "" {
 		log.Println("Searching for schema by CIR Instrument ID: " + cirInstrumentId)
--- a/templates/launch.html
+++ b/templates/launch.html
@@ -2,604 +2,604 @@
 
 {{ define "body" }}
 
-<body onload="onLoad()">
-    <h1>Launch a survey</h1>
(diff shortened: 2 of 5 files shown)


rmccar:
[Launching via schema url via Launcher raises error in Runner](https://trello.com/c/XA4m5xjI/5980-launching-via-schema-url-via-launcher-raises-error-in-runner)

templates/launch.html
berroar, line 1: Hard to tell what's changed re the quicklauncher with so many other changes in this PR now - particularly this file. I would split out the tidying of the code to another PR so it doesn't muddy the waters of the actual change, and best practice as it makes it easier for anyone who needs to go back through the commit history to see what changed in future.

katie-gardner (changes requested):
Works for v2 launching 👍 
Doesn't work for v1, the launch errors because runner is expecting eq_id and form_type or the schema name, but in this case it doesn't have it

I think possibly the change for the v1 launch pattern should be undone? Since the validation in runner for v1 is ok with it the way it was, just v2 that doesn't work

templates/launch.html
katie-gardner: Guessing it was an autoformat, all whitespace changes, do we have a card for using prettier or something? I feel it'd be useful but I recall us discussing this and deciding against it I think 🤔

rmccar: So I've removed a load of whitespace, that should be easy to see but the main change that has caused that is that I've changed some of the indentation. I can revert that particular change though and move that into another PR. Just removing that change would make it easier to see the changes I think.

katie-gardner: I reckon it's probably ok, the 'Hide whitespace' option on the review files omits that indentation/removal so can review it OK

rmccar: Ah yeah just tried that, makes it much easier to see the changes 👍

rmccar:
> Works for v2 launching 👍 Doesn't work for v1, the launch errors because runner is expecting eq_id and form_type or the schema name, but in this case it doesn't have it
> 
> I think possibly the change for the v1 launch pattern should be undone? Since the validation in runner for v1 is ok with it the way it was, just v2 that doesn't work

Fixed this now 👍

katie-gardner (approved):
All seems to be working for me 👍

berroar:
> > Works for v2 launching 👍 Doesn't work for v1, the launch errors because runner is expecting eq_id and form_type or the schema name, but in this case it doesn't have it
> > I think possibly the change for the v1 launch pattern should be undone? Since the validation in runner for v1 is ok with it the way it was, just v2 that doesn't work
> 
> Fixed this now 👍

I still seem to get this error, when launching a test schema by URL in v1 (was using test_checkbox) 🤔

surveys/surveys.go
```diff
@@ -281,8 +281,7 @@ func GetLauncherSchema(schemaName string, schemaUrl string, cirInstrumentId stri
 	if schemaUrl != "" {
 		log.Println("Getting schema by URL: " + schemaUrl)
 		launcherSchema = LauncherSchema{
-			URL:  schemaUrl,
-			Name: schemaName,
+			URL: schemaUrl,
```
katie-gardner, line 284: I'm not sure why intermittently had success with v1 launcher, but it looks like this is the issue for the v1 launch - because `schemaName` isn't populated anymore, the block of code that you put back isn't going to set the name, which is why v1 launch is still missing it

rmccar: Yeah I thought this would be the issue if there was one. I was waiting to have a chat with you guys before removing this 👍

berroar (approved):
Tested, LGTM 👍

katie-gardner (approved):
All working as expected now 👍

PR #64, opened 2024-01-08
Fix improper survey metadata generation when using booleans

VirajP1002:
Previously `test_metadata_routing.json` was failing to open because `flag_1` (now changed to `boolean_flag`) wasn't being included within `survey_metadata` however it was included as top level metadata leading. 

In order to fix this issue, the logic in `getSurveyMetadataFromClaims` was flipped where metadata in `claimValues` that is not top level metadata is considered to be survey metadata.

`claimValues` used in `getSurveyMetadataFromClaims` did not contain `boolean_flag` when unticked in launcher (set to False) and so the logic where `boolean_flag` is added back to `survey_metadata` for token generation in `GenerateTokenFromPost` had to be changed as the previous implementation added `boolean_flag` into top level rather then in `survey_metadata`.

`launch.html` was updated to ensure that the form submission doesn't contain unwanted metadata.

### How to review
- Ensure the changes made are suitable

- Run Launcher and Runner locally and ensure `test_metadata_routing.json` opens when `boolean_flag` is set to True/False in the Launcher UI.

- Ensure `boolean_flag` is not put in top level metadata, do this by adding a breakpoint to line 110 in `sessions.py` in Runner, open launcher and open `test_metadata_routing.json`. In the debug window, select `decrypted_token` and check that `boolean_flag` is within `survey_metadata` and is the correct value.

- Check the loaded values on the launcher UI are correct, do this by adding metadata to a schema such as `{
            "name": "test",
            "type": "date"
        }`. Load up the launcher UI and check that it has been added in the `Survey Metadata` section with a value in the field. The same can be checked for altering the type to `string`,`uuid` or `url`. If you run Runner in debug mode, you can check the token to ensure `test` in within `survey_metadata`.

- Check `test_metadata_routing` opens in quick-launch, add an additional piece of metadata (similar to the one above) to the schema and run Runner in debug mode. Enter the url into the browser and check that the survey_metadata is correct, it should match the one from the UI. You can also add additional arguments to the URL or override an existing piece of metadata and ensure the `survey_metadata` section is correct.

- Ensure changes do not break other schemas, e.g. test_supplementary_data, or any business/social surveys


berroar (changes requested):
Looks good so far, but as per the ticket we need to implement this part as well: `If a key type is unknown, and therefore launcher doesn’t have a default for it, use the type of the metadata value to auto-generate something suitable`. At the moment, for strings Launcher is sending an empty string e.g. "", but if you added a new metadata with a type like `date` to metadata in your test schema, the launch would fail as there would be no value set (as no default exists). So we need to dynamically generate values based on the type

authentication/auth.go
```diff
@@ -250,7 +219,29 @@ func generateClaimsV2(claimValues map[string][]string, launcherSchema surveys.La
...
+	schema.Metadata = append(mandatoryClaims, schema.Metadata...)
+	for _, mandatoryMetadata := range mandatoryClaims {
+		present := false
+		for _, schemaMetadata := range schema.Metadata {
+			if mandatoryMetadata.Name == schemaMetadata.Name {
+				present = true
+				break
+			}
+		}
+		schema.Metadata = append(schema.Metadata, mandatoryMetadata)
```
berroar, line 238: Slightly confused about the logic here - as we are appending to the metadata if it is present or not?

VirajP1002: I was also slightly confused to my logic here 😆 I've updated it and hopefully it makes more sense.

authentication/auth.go
```diff
@@ -527,16 +505,24 @@ func GenerateTokenFromDefaultsV2(schemaURL string, accountServiceURL string, url
...
 	initialData := surveyMetadata["data"].(map[string]interface{})
 
 	for key, value := range initialData {
 		updatedData[key] = value
 	}
 
+	requiredSchemaMetadata = addAdditionalMetadata(surveyMetadata, requiredSchemaMetadata)
+
+	for _, metadata := range requiredSchemaMetadata {
+		if metadata.Validator == "boolean" {
```
petechd, line 518: Are these defaults not generated (is this method not used) only if you go through quick launch? It seems like it's only executed inside this `quickLauncherHandler()` function.

authentication/auth.go
```diff
@@ -558,6 +544,21 @@ func GenerateTokenFromDefaultsV2(schemaURL string, accountServiceURL string, url
...
 }
 
+func addAdditionalMetadata(surveyMetadata map[string]interface{}, requiredSchemaMetadata []Metadata) []Metadata {
+	for _, metadataVal := range surveyMetadata {
+		for key, value := range metadataVal.(map[string]interface{}) {
+			newMetadata := Metadata{}
+			if strings.Contains(value.(string), "True") || strings.Contains(value.(string), "False") {
+				newMetadata = Metadata{Name: key, Validator: "boolean", Default: "false"}
+			} else {
+				newMetadata = Metadata{Name: key, Validator: "string", Default: "test"}
```
petechd, line 554:
```suggestion
				newMetadata = Metadata{Name: key, Validator: "string", Default: value.(string)}
```

VirajP1002: Updated 👍

authentication/auth.go
```diff
@@ -245,12 +218,32 @@ func generateClaimsV2(claimValues map[string][]string, launcherSchema surveys.La
 	surveyMetadata := make(map[string]interface{})
 	data := make(map[string]interface{})
 
-	if schema.SurveyType == "social" || schema.SurveyType == "health" {
+	if isNonBusinessNonDefaultSurvey(schema.SurveyType) {
```
berroar, line 221: Rather than NonBusiness we can just use Social, but what is meant by NonDefault?

authentication/auth.go
```diff
@@ -527,16 +509,32 @@ func GenerateTokenFromDefaultsV2(schemaURL string, accountServiceURL string, url
...
 
+	/*
+		This method below is used to add any metadata (that is a parameter) which is intended to be of type boolean to
+		requiredSchemaMetadata for the typing in "updatedData" to be correct. Without this method, a parameter added to
+		the URL set as a boolean for example, "&example=True", would be interpreted as a String hence the type in the
+		token would be incorrect. It doesn't apply to other types because those do not need to be updated/changed in
+		"updatedData". It's essential to ensure the values in "updatedData" are of the correct type for the types in
+		token to be correct.
+	*/
+	requiredSchemaMetadata = addAdditionalMetadata(surveyMetadata, requiredSchemaMetadata)
```
berroar, line 527: Is this only for quick launch? Does the method need to be renamed if it is only handling bools to make it clearer what is being done?

authentication/auth.go
```diff
@@ -558,6 +556,32 @@ func GenerateTokenFromDefaultsV2(schemaURL string, accountServiceURL string, url
 	return token, ""
 }
 
+func addAdditionalMetadata(surveyMetadata map[string]interface{}, requiredSchemaMetadata []Metadata) []Metadata {
```
berroar, line 559: Do we need the nested loops here if we are just updating the map with any quick launch bool values?

authentication/auth.go
```diff
@@ -558,6 +556,32 @@ func GenerateTokenFromDefaultsV2(schemaURL string, accountServiceURL string, url
...
+	for _, metadataVal := range surveyMetadata {
+		for key, value := range metadataVal.(map[string]interface{}) {
+			if isExistingRequiredSchemaMetadata(requiredSchemaMetadata, key) {
+				continue
+			} else {
+				newMetadata := Metadata{}
```
berroar, line 565: Don't think we need this line, can just use `newMetadata := Metadata{Name: key, Validator: "boolean", Default: "false"}` below?

authentication/auth.go
```diff
@@ -558,6 +556,32 @@ func GenerateTokenFromDefaultsV2(schemaURL string, accountServiceURL string, url
...
+				if strings.Contains(value.(string), "True") || strings.Contains(value.(string), "False") {
```
berroar, line 566: How come we're expecting the URL param to be `True` or `False` when the value is `true/false` - should we handle both in URL?

authentication/auth.go
```diff
@@ -122,41 +122,11 @@ type Metadata struct {
...
-		"sds_dataset_id",
-		"WINDOW_START_DATE",
-		"WINDOW_CLOSE_DATE",
-		"PORTAL_ID",
-		"PARTICIPANT_WINDOW_ID":
-
-		return true
-	}
-	return false
+func (metadata *Metadata) updateMetadataDefault(newDefault string) {
```
berroar, line 125: I don't think this method is being used at the moment?

authentication/auth.go
```diff
@@ -527,16 +509,32 @@ func GenerateTokenFromDefaultsV2(schemaURL string, accountServiceURL string, url
...
 
 	initialData := surveyMetadata["data"].(map[string]interface{})
 
 	for key, value := range initialData {
 		updatedData[key] = value
 	}
 
+	/*
+		This method below is used to add any metadata (that is a parameter) which is intended to be of type boolean to
+		requiredSchemaMetadata for the typing in "updatedData" to be correct. Without this method, a parameter added to
```
berroar, line 521: Maybe I'm misunderstanding but `requiredSchemaMetadata` should only refer to metadata that is specified in the actual schema, I don't think we need to add any values to it?

authentication/auth.go
VirajP1002: I created it while implementing a solution to removing duplicate metadata items. I left it in because it could be used in the future. But since it's not used there may be no point and can be recreated easily if necessary, I removed it 👍

authentication/auth.go
VirajP1002: My reasoning for `True` and `False` was to follow the Python convention for `True/False` with capital letters, so if someone wanted to have a parameter set to a boolean they would follow this standard. If we handle both, there's still the chance that someone could intend for 'true' or 'false' to be a string rather than a boolean.

authentication/auth.go
VirajP1002: When I created this method, I thought it would be suitable as NonBusiness as the method checks if the type is `social`, `health` or `ukhsa-ons`. I thought NonDefault meant anything that wasn't considered the three I mentioned.

authentication/auth.go
VirajP1002: This is only for quick launch. A more appropriate name could be `addUrlBooleanMetadata`?

authentication/auth.go
berroar: Those types are all classed as Social Surveys though (even Health) in ONS terms.

authentication/auth.go
MebinAbraham: The standard is to support all toggle words, `on`, `off` etc. But here I would say handle it without case-sensitive, so compare the equal case so that it handled both upper and lower case.

MebinAbraham:
Not directed at your changes, but this shows how much of a mess the current Go logic is. Unnecessary complexity and confusing logic in quite a few places. Even though it is a dev tool, it might be worth us doing a refactor.

VirajP1002:
> Not directed at your changes, but this highlights how much of a mess the current Go logic is. Unnecessary complexity and confusing logic in quite a few places. Even though it is a dev tool, it might be worth us doing a refactor.

I would say it's quite messy by the fact it's hard to navigate, for example going to method calls and definitions is odd as it jumps around the page and that some v1 methods are still there.

authentication/auth.go
VirajP1002: Updated in new PR

authentication/auth.go
VirajP1002: Updated in new PR

authentication/auth.go
VirajP1002: No problem, will handle in the new PR

authentication/auth.go
```diff
@@ -245,12 +214,32 @@ func generateClaimsV2(claimValues map[string][]string, launcherSchema surveys.La
...
 	data := make(map[string]interface{})
 
-	if schema.SurveyType == "social" || schema.SurveyType == "health" {
+	if isNonBusinessNonDefaultSurvey(schema.SurveyType) {
 		receiptingKeys := []string{"qid"}
 		surveyMetadata["receipting_keys"] = receiptingKeys
 	}
 
-	getSurveyMetadataFromClaims(claimValues, data, claims, surveyMetadata)
+	/*
```
berroar, line 222: I'm slightly confused about the purpose of this block, as when I remove it the behaviour seems to be the same, and I don't get any duplicates? Is there are specific scenario this was added for?

VirajP1002: It was added it because I thought that `ru_ref` wasn't being handled and would have been placed in the incorrect level for both quick launch and by the UI, seems like it was already being handled else where and should be fine to remove. Would you mind verifying if ru_ref appears in the correct level in the token?

authentication/auth.go
VirajP1002: Updated 👍

launch.go
```diff
@@ -170,11 +170,11 @@ func quickLauncherHandler(w http.ResponseWriter, r *http.Request) {
 	AccountServiceLogOutURL := getAccountServiceURL(r)
 	urlValues := r.URL.Query()
 	schemaURL := urlValues.Get("url")
+	delete(urlValues, "url")
```
berroar, line 173: How come we're deleting this?

VirajP1002: `schema_url` in top level is the same value so when we do not have this in, the `url` value is passed to auth.go and is then added as survey metadata leading to urls appearing in both top level and survey metadata, we only want one and it has to be in top level according the ONS schema definitions

VirajP1002: Screenshot 2024-02-02 at 14 23 12
This is what the token would appear as without that line in place.

berroar: I guess it shouldn't be added in the first place rather than deleting it though?

VirajP1002: It is needed when `GenerateTokenFromDefaultsV2` is called in `launch.go` because it is a parameter. 

I did notice we use a separate method called `getSchemaClaims` which is used to add in `schema_url` in to top level metadata in `GenerateTokenFromDefaultsV2` and is defined in `auth.go`. 

One alternative I thought of was rather than deleting it, it's possible to add an `if` statement in `getSurveyMetadataFromClaims` to ignore a url value. Another thought was to rename the variable from `url` to `schema_url` within `urlValues` and then stop the use of `getSchemaClaims` but I think this wouldn't work for cir as it needs `getSchemaClaims` for it to be in the token. I believe it would have to be deleted/ignored to prevent duplication in the token.

berroar: I see this is only for quick launch, so I think maybe `url` we should be made consistent with the actual `schema_url` field in launcher (unless there is a reason why these are different 🤔)?

VirajP1002: As discussed, I changed the query to `schemaURL := urlValues.Get("schema_url")` and updated the README 👍

petechd (approved):
All the issues above seems to be resolved and all the functionality (scenarios) mentioned work as expected, code is hard to follow but not because of PR creator's design.

authentication/auth.go
```diff
@@ -793,8 +780,8 @@ func GetDefaultValues() map[string]string {
 	defaults["TEST_QUESTIONS"] = "F"
 	defaults["sds_dataset_id"] = sdsDatasetId.String()
 	defaults["survey_id"] = "123"
-	defaults["WINDOW_START_DATE"] = "2023-03-01"
-	defaults["WINDOW_CLOSE_DATE"] = "2023-03-31"
+	defaults["WINDOW_START_DATE"] = "2016-05-01"
```
petechd, line 783: Why was this date changed?

VirajP1002: I changed this date for it to be consistent with the `ref_p_start_date` and `ref_p_end_date` as Mebin and I discussed this too.

templates/launch.html
```diff
@@ -164,12 +164,12 @@ <h3>Runner Data</h3>
 
             <div class="field-container">
                 <label for="account_service_log_out_url">Account Service Log Out URL</label>
-                <input id="account_service_log_out_url" name="account_service_log_out_url" type="text" value="{{.AccountServiceLogOutURL}}" class="qa-account_service_log_out_url">
+                <input id="account_service_log_out_url" type="text" value="{{.AccountServiceLogOutURL}}" class="qa-account_service_log_out_url">
```
berroar, line 167: How come we don't send this one?

VirajP1002: We didn't send it because it's not included in the ons-schema-definitions doc, so I removed it because I didn't think it needed to be in the token

VirajP1002: In fact, it's included as an optional field, but that's only in the v1 payload document not in the v2 payload doc, so since we are planning to remove v1, I guess this would be appropriate?

berroar: Hmm it should be removed when the ticket to remove v1 gets picked up, but seems like the type of thing that could be easily missed

berroar (approved):

PR #65, opened 2024-02-01
Ensure boolean parameters in quick launch are typed correctly

VirajP1002:
When working on https://github.com/ONSdigital/eq-questionnaire-launcher/pull/64 I noticed that  when using quick launch, adding a parameter  to a URL which is intended to be a boolean would be a string in the token (incorrect type). In order to correct this, I had to add a method in to create a metadata object and add it to requiredSchemaMetadata in order for it to be casted to a boolean (which is done by the logic that is already in place).

(Note: When reviewing, the changes made is the refactoring of the solution I used to this problem in https://github.com/ONSdigital/eq-questionnaire-launcher/pull/64.)

### How to review
- Check the changes made are appropriate
- Run Runner in debug mode and add a breakpoint in line 110 in `session.py`. Then use quick launch and add a parameter to the URL such as `test=false` or `test=true` and see if the type in the token. (Note: The values of true and false are case insensitive so using `test=True` for example, will still result in a boolean in the token).


VirajP1002:
[Bug: Launcher improperly generating survey metadata](https://trello.com/c/XSA1hLNP/5953-bug-launcher-improperly-generating-survey-metadata)

authentication/auth.go
```diff
@@ -556,17 +556,13 @@ func GenerateTokenFromDefaultsV2(schemaURL string, accountServiceURL string, url
...
-			} else {
-				newMetadata := Metadata{}
-				if strings.Contains(value.(string), "True") || strings.Contains(value.(string), "False") {
-					newMetadata = Metadata{Name: key, Validator: "boolean", Default: "false"}
-					requiredSchemaMetadata = append(requiredSchemaMetadata, newMetadata)
-				}
+func addUrlBooleanMetadata(updatedMetadata map[string]interface{}, requiredSchemaMetadata []Metadata) []Metadata {
+	for key, value := range updatedMetadata {
+		if !(isExistingRequiredSchemaMetadata(requiredSchemaMetadata, key)) {
+			value = strings.ToLower(value.(string))
```
petechd, line 562: If you create var called `value` once, used it in both cases on line 563 as well.

authentication/auth.go
```diff
@@ -556,17 +556,13 @@ func GenerateTokenFromDefaultsV2(schemaURL string, accountServiceURL string, url
...
+			if strings.Contains(value.(string), "true") || strings.Contains(value.(string), "false") {
```
petechd, line 563: Not only it contains but I would say it equals, so use the equals method in Go, something like:
```go
var testedString = strings.ToLower(value.(string))
strings.EqualFold(testedString, "true")
```

authentication/auth.go
petechd, line 562: You are reassigning `value` var from the top, which is bad practice, create a new var, so something like:
```go
for key, value := range updatedMetadata {
		if !(isExistingRequiredSchemaMetadata(requiredSchemaMetadata, key)) {
			var comparedValue = strings.ToLower(value.(string))
```

petechd: And then reuse the new var in your:
```go
if strings.Contains(value.(string), "true") || strings.Contains(value.(string), "false")
```

authentication/auth.go
VirajP1002: Nice spot! Contains could have lead to some issues further on, updated 👍

authentication/auth.go
VirajP1002: Updated 👍

authentication/auth.go
```diff
@@ -492,6 +492,16 @@ func GenerateTokenFromDefaultsV2(schemaURL string, accountServiceURL string, url
 		updatedData[key] = value
 	}
 
+	/*
```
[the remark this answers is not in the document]
VirajP1002: Makes sense, just because of the confusing layout, I thought it would make sense to have a longer explanation with context, but it may be a bit too long 😅. I've made the comment smaller 👍

authentication/auth.go
```diff
@@ -522,6 +529,28 @@ func GenerateTokenFromDefaultsV2(schemaURL string, accountServiceURL string, url
 	return token, ""
 }
 
+func addUrlBooleanMetadata(updatedMetadata map[string]interface{}, requiredSchemaMetadata []Metadata) []Metadata {
+	for metadataName, metadataValue := range updatedMetadata {
+		if !(isExistingRequiredSchemaMetadata(requiredSchemaMetadata, metadataName)) {
```
berroar, line 534: I think this if statement can be removed which will reduce the complexity here, it looks like the duplicate value that would be generated is already being handled and overwritten elsewhere. The token generated looks fine in runner without this 👍

VirajP1002: Makes sense, updated 👍

authentication/auth.go
```diff
@@ -522,6 +528,17 @@ func GenerateTokenFromDefaultsV2(schemaURL string, accountServiceURL string, url
 	return token, ""
 }
 
+func addUrlBooleanMetadata(updatedMetadata map[string]interface{}, requiredSchemaMetadata []Metadata) []Metadata {
+	for metadataName, metadataValue := range updatedMetadata {
+		convertedValue := strings.ToLower(metadataValue.(string))
+		if strings.EqualFold(convertedValue, "true") || strings.Contains(convertedValue, "false") {
+			newMetadata := Metadata{Name: metadataName, Validator: "boolean", Default: "false"}
```
berroar, line 535: Minor, but don't think we need a var for this since it's only used once

berroar (approved):

petechd (approved):

PR #69, opened 2024-03-27
Support business prepop survey launch

liamtoozer:
To allow launching of business prepop surveys, our Launcher service needs to be altered to handle:

* Displaying of SDS `dataset_id` either in survey metadata (if present) or in the SDS section
* Sending the `survey_id` to Runner if it's a business survey only if the `survey_id` is not present in the survey metadata

### How to review
Run with the Mock SDS branch. Ensure that all relevant prepop surveys launch correctly and display the prepop data in Runner.

There might be other launch scenarios that I've missed, so would be grateful for any advice if so

Jira: https://jira.ons.gov.uk/browse/ECI-1122


liamtoozer:
Will re-raise

PR #71, opened 2024-04-05
Remove ability to launch using the v1 pattern

berroar:
Removes the ability to launch surveys using the v1 format. v1 was deprecated but support in runner is now being removed so we will no longer need to launch using this pattern.

### How to review
Check references to the v1 launch pattern have been removed.
Check that you are still able to launch surveys using the v2 pattern.


launch.go
```diff
@@ -167,20 +163,11 @@ func redirectURL(w http.ResponseWriter, r *http.Request) {
 func quickLauncherHandler(w http.ResponseWriter, r *http.Request) {
 	hostURL := settings.Get("SURVEY_RUNNER_URL")
 	accountServiceURL := getAccountServiceURL(r)
-	AccountServiceLogOutURL := getAccountServiceURL(r)
 	urlValues := r.URL.Query()
 	schemaURL := urlValues.Get("schema_url")
-	version := urlValues.Get("version")
```
liamtoozer, line 173: I'm not sure if it's still in use, but there's [a bit in the readme](https://github.com/ONSdigital/eq-questionnaire-launcher#:~:text=The%20optional%20query%20parameter%20version%20can%20be%20added%20to%20the%20quick%20launch%20url%20which%20allows%20for%20the%20launch%20payload%20structure%20to%20be%20specified.) about quick-launch with an optional `version` in the query string. Would that need to be amended/removed too?

liamtoozer (approved):
Works great for me 👍

launch.go
berroar: We still want the ability to support being able to set other versions, even if we only support v2 for now, so I guess this can stay

VirajP1002 (approved):
I didn't see any mentions to the v1 pattern and I opened several schemas successfully!

PR #74, opened 2024-04-29
Display SDS dataset IDs either with survey metadata or in SDS section

liamtoozer:
To allow launching of business prepop surveys, our Launcher service needs to be altered to handle:

* Displaying of SDS `dataset_id` either in survey metadata (if present) or in the SDS section
* Sending the `survey_id` to Runner if it's a business survey only if the `survey_id` is not present in the survey metadata

### How to review
Run with the Mock SDS branch. Ensure that all relevant prepop surveys launch correctly and display the prepop data in Runner.

There might be other launch scenarios that I've missed, so would be grateful for any advice if so

Jira: https://jira.ons.gov.uk/browse/ECI-1122

### DIFF
--- a/static/javascript/launch.js
+++ b/static/javascript/launch.js
@@ -139,6 +139,9 @@
 // store fetch so it only needs to be re-done if the survey changes
 let supplementaryDataSets = null;
 
+// We always need survey_id from top-level schema metadata for SDS retrieval
+let schemaSurveyId = null;
+
 function clearSurveyMetadataFields() {
   document.querySelector("#survey_metadata_fields").innerHTML = "";
   showSupplementaryData(false);
@@ -366,7 +369,6 @@ async function loadSDSDatasetMetadata(survey_id, period_id) {
 
 function handleNoSupplementaryData() {
   showSupplementaryData(false);
-  showSubmitFlushButtons(false);
 }
 
 function showCIRMetdata(cirInstrumentId, cirSchema) {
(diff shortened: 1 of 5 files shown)


berroar (approved):
Able to launch, LGTM 👍

static/javascript/launch.js
```diff
@@ -366,7 +369,6 @@ async function loadSDSDatasetMetadata(survey_id, period_id) {
 
 function handleNoSupplementaryData() {
   showSupplementaryData(false);
-  showSubmitFlushButtons(false);
```
petechd, line 369: Why was this removed? There is something odd happening now when I try to flush the data, the flush button opens a schema but it seems to be happening in staging as well so maybe just some bug.

liamtoozer: Had a quick discussion about this but just to confirm - I wanted to guard against `404` responses from the mock SDS service, as we're now calling that API **every** time for **any** schema now. However, it looks like `404`s are [gracefully handled by te backend](https://github.com/ONSdigital/eq-questionnaire-launcher/blob/6d17e39e6234fdcf282d741951d429a453977cc1/surveys/surveys.go#L261) and return an empty list with a `200` status code to the client.

Will add this back in so that any response from the mock SDS that's not a `200` or `404` will cause a failure and will therefore hide the buttons 👍

petechd (approved):

PR #70, opened 2024-03-27
Add Formatting and Linting to Launcher

VirajP1002:
### What is the context of this PR?
Describe what you have changed and why, link to other PRs or Issues as appropriate.

### How to review
Describe the steps required to test the changes (include screenshots if appropriate).

PR #77, opened 2024-05-07
Add action_flush name to flush survey data button

petechd:
This reintroduces the form name "action_flush" so our logic (referring it) works as expected. It was removed in [formatting and linting PR](https://github.com/ONSdigital/eq-questionnaire-launcher/pull/70).

### How to review
To test flush survey data working you need to:
- open a schema using launcher
- answer few questions
- before submitting (the whole survey) go back to the launcher screen
- load a previous value for response id field and pick the same schema name
- **Important**: choose the flusher role
- press the flush survey data button
You should hit the flush endpoint and get status code 200.
In other scenarios you should end up on the /flush endpoint and get different statuses. Just follow the logic below.
The endpoint logic in runner:
```python
@flush_blueprint.route("/flush", methods=["POST"])
def flush_data() -> Response:
    if session:
        session.clear()

    encrypted_token = request.args.get("token")

    if not encrypted_token or encrypted_token is None:
        return Response(status=403)

    decrypted_token = decrypt(
        token=encrypted_token,
        key_store=_get_keystore(),
        key_purpose=KEY_PURPOSE_AUTHENTICATION,
        leeway=current_app.config["EQ_JWT_LEEWAY_IN_SECONDS"],
    )

    roles = decrypted_token.get("roles")

    if roles and "flusher" in roles:
        user = _get_user(decrypted_token["response_id"])

        if metadata := get_metadata(user):
            contextvars.bind_contextvars(
                tx_id=metadata.tx_id,
                ce_id=metadata.collection_exercise_sid,
            )
            bind_contextvars_schema_from_metadata(metadata)

        if _submit_data(user):
            return Response(status=200)
        return Response(status=404)
    return Response(status=403)
```
the logic in launcher:
```go
flushAction := r.PostForm.Get("action_flush")
	log.Println("Request: " + r.PostForm.Encode())

	if flushAction != "" {
		http.Redirect(w, r, hostURL+"/flush?token="+token, 307)
	} else if launchVersion != "" {
		http.Redirect(w, r, hostURL+"/session?token="+token, 301)
	} else {
		http.Error(w, fmt.Sprintf("Invalid Action"), 500)
	}
```
To make results more obvious you can modify the runner code to something like:
```python
        if _submit_data(user):
            return "<h1>Successfully flushed</h1>"
        return Response(status=404)
    return Response(status=403)
```


rmccar (approved):

Yuyuutsu (approved):
Works as expected 👍

PR #82, opened 2024-06-07
Update launcher to use the Design System

VirajP1002:
To update the front end of Launcher, we decided to use DS components. However, given the fact that Launcher in written in Go, using the macros wasn't possible so the HTML had to be hardcoded within the templates files and the components were imported using the CDN link.

The changes to `launch.html` include creating to separate columns in order for the right side of the page to be utilised. The first column includes the options for selecting a schema either by name or using CIR. The second column is the metadata section, it uses the accordion component which dynamically updates the metadata sections so only the relevant metadata appears. For example, if you choose a business schema it shows the business metadata, survey metadata, required metadata and runner metadata sections and if you select a test schema it removes the business metadata section and updates the remaining sections. The accordion also has a save state so if you opened a specific section, opened a survey and went back to launcher (only using the back arrow), it would remain open.

The changes to `layout.html` includes a CDN link for the `main.css` file. The changes to `static/css/main.css` includes an update of the background colour of Launcher and removing any redundant CSS classes as the DS handles this. The changes to `launch.js` includes updates to support the accordion and to remove any redundant JS.

(Note: The accordion elements are hidden rather than injected in via JS because the DS adds attributes to the accordion elements on page load and the JS would be injected after page load hence the attributes needed won't be present).

### How to review
- _**Make sure to clear your browser cache before reviewing**_
- Check the changes made to all the CSS/JS/HTML files
- Check the formatting of the files
- Ensure the functionality of Launcher is still the same and everything works as intended
- Check if any redundant CSS/JS/HTML has been missed.


templates/launch.html
```diff
@@ -1,215 +1,395 @@
...
+                    <div class="ons-u-mt-s">
+                        <button id="schema-url-btn"
+                                type="button"
+                                class="ons-btn qa-btn-submit-dev"
+                                onclick="loadMetadataForRemoteSchema()">
+                            <span class="ons-btn__inner"><span class="ons-btn__text">Load Metadata</span></span>
+                        </button>
+                    </div>
+                    <hr>
+                    <h3 class="ons-u-mt-m">Survey Actions</h3>
```
rmccar, line 89: This one should also be a `<h2>` because it comes under the `<h1>` in the hierarchy

static/javascript/launch.js
```diff
@@ -354,9 +358,12 @@ function getInputField(
   const value = defaultValue ? `value="${defaultValue}"` : "";
   const readOnly = isReadOnly ? "readonly" : "";
   if (readOnly) {
-    return `<input ${readOnly} id="${fieldName}" type="${type}" ${value} class="qa-${fieldName}" onchange="${onChangeCallback}">`;
+    return `<input ${readOnly} id="${fieldName}" type="${type}" ${value} class="qa-${fieldName} ons-input ons-input--text" onchange="${onChangeCallback}">`;
```
rmccar, line 361: Are these "qa" classes doing anything? Ive searched the repo and they only seem to be set in some places in the js but not actually used for anything

rmccar (changes requested):
Noticed these small issues but also theres a tabbing issue where because of the way you are adding details components when you tab through the inputs you are able to focus on the details components which is confusing because they aren't there yet. This is because there is JS from the DS adding `tabindex=0` to those elements when they are initialised on page load. Not a big issue and maybe something we can live with or introduce some js to add and remove the tab index when those are added/removed.

The other issue is when using this on smaller screens. There are some issues with the layout. I know this is a dev tool so maybe something we can live with but some of these should be fairly easy to fix especially the buttons at the bottom. Should be able to add a `margin-botton` to them, remove the `margin-left` and add a `margin-right`. Would be nice to fix the schema select which is overflowing too.

Screenshot 2024-06-11 at 10 39 49

petechd:
Not sure if I am doing something wrong but it seems like the functionalities are not the same as before, Load Schema button is disabled after I pick the schema name from the dropdown.

static/javascript/launch.js
```diff
@@ -584,9 +591,11 @@ function retrieveResponseId() {
   let responseIdButton = document.querySelector("#response-id-btn");
 
   if (responseId) {
-    responseIdButton.classList.remove("btn--hidden");
+    responseIdButton.classList.remove("ons-btn--disabled");
+    responseIdButton.removeAttribute("disabled");
   } else {
-    responseIdButton.classList.add("btn--hidden");
+    responseIdButton.classList.add("ons-btn--disabled");
+    responseIdButton.setAttribute("disabled");
```
petechd, line 598: This cannot be done using `setAttribute()` as it always requires two attributes, name and value. Consider using `disabled = boolean` as described in: https://www.w3schools.com/jsref/prop_pushbutton_disabled.asp.

static/javascript/launch.js
VirajP1002: I've had a look now and the places they are referred to in JS is just to inject HTML back into the code, maybe they can be removed?

templates/launch.html
```diff
@@ -1,215 +1,395 @@
...
-            <input type="button"
-                   value="Clear Local Storage"
-                   class="qa-btn-submit-dev btn"
-                   id="local-storage-btn"
-                   onclick="clearLocalStorage()" />
+        <div class="ons-grid">
+            <div class="ons-grid__col ons-col-6@m ">
+                <div class="ons-pl-grid-col">
+                    <h1>Launch a survey</h1>
+                    <strong class="ons-u-mt-m"><u>Launch Pattern</u></strong>
```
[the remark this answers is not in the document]
VirajP1002: Slightly unsure about this change as it looks a bit strange, could we discuss in a huddle?

static/javascript/launch.js
rmccar: Yeah, remove them and give it a good test

VirajP1002:
> Noticed these small issues but also theres a tabbing issue where because of the way you are adding details components when you tab through the inputs you are able to focus on the details components which is confusing because they aren't there yet. This is because there is JS from the DS adding `tabindex=0` to those elements when they are initialised on page load. Not a big issue and maybe something we can live with or introduce some js to add and remove the tab index when those are added/removed.
> 
> The other issue is when using this on smaller screens. There are some issues with the layout. I know this is a dev tool so maybe something we can live with but some of these should be fairly easy to fix especially the buttons at the bottom. Should be able to add a `margin-botton` to them, remove the `margin-left` and add a `margin-right`. Would be nice to fix the schema select which is overflowing too.
> 
> Screenshot 2024-06-11 at 10 39 49

I've seemed to address this issue in both columns, although the size of the required metadata labels are smaller, they don't overflow

petechd:
Worth double checking, it seems like  the `Flush Survey Data` button in your PR stopped working and it opens a schema instead of flushing now?

VirajP1002:
> Worth double checking, it seems like the `Flush Survey Data` button in your PR stopped working and it opens a schema instead of flushing now?

Yep, a missing `value` attribute in the button was impacting the logic used in launch.go for flushing, fixed this now 👍

static/javascript/launch.js
VirajP1002: Updated 👍

static/javascript/launch.js
VirajP1002: updated 👍

templates/launch.html
VirajP1002: updated 👍

templates/launch.html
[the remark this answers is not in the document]
VirajP1002: As discussed in the huddle we spoke about the reasoning and what DS classes could be used, updated 👍

templates/launch.html
```diff
@@ -1,215 +1,405 @@
...
-                   value="Flush Survey Data"
-                   class="qa-btn-submit-dev btn btn--hidden"
-                   id="flush-btn" />
-            <input type="button"
-                   value="Clear Local Storage"
-                   class="qa-btn-submit-dev btn"
-                   id="local-storage-btn"
-                   onclick="clearLocalStorage()" />
+        <div class="ons-grid">
+            <div class="ons-grid__col ons-col-6@m ">
```
rmccar, line 9:
```suggestion
            <div class="ons-grid__col ons-col-6@l ">
```
I think we talked about this but I think it could be nice if the panels went under each other at a larger breakpoint. Line 117 will also need the same change. Try this and see what you think. I think its an improvement.

Yuyuutsu (commented):
Would it be possible if the `Show all` button be the same size as the other buttons. More of a personal prefrence. I dont mind if everyone else wants to keep it as it is 👍

templates/launch.html
VirajP1002: I made that change I used `@xl` instead as it makes the buttons appear less squished, it looks better than before

VirajP1002:
> Would it be possible if the `Show all` button be the same size as the other buttons. More of a personal prefrence. I dont mind if everyone else wants to keep it as it is 👍

As discussed the text within the button is determines the size of it, happy to change the text to something like `Show Metadata`. We can see what the other devs think too 👍

rmccar:
Screenshot 2024-06-24 at 11 17 42

Theres a bit of extra space added on here when focused on this details component. It only seems to happen on the health or business ones

VirajP1002:
> Screenshot 2024-06-24 at 11 17 42
> 
> Theres a bit of extra space added on here when focused on this details component. It only seems to happen on the health or business ones

Nice spot, I've address this now, when I was injecting the title in via the JS, it was previously adding in the title within h2 tags, so I just made the JS insert the string instead which fixed the issue 👍

rmccar:
Just checked your latest and noticed a pretty minor thing. You have a margin left on the left of these 3 buttons which is pushing them over a bit and arent aligned because of it.

Screenshot 2024-06-25 at 10 52 55

The thing is, If you just remove this it causes this problem:

Screenshot 2024-06-25 at 10 53 08

So what you need to do is if these have a container element add `display: flex`, flex-wrap: wrap` and something like `gap: 1rem` to it in CSS. Then remove all the margins. This should solve that for you.

liamtoozer (commented):
Really nice - great work! It looks loads better than before 😄 Just a few very minor cosmetic things I spotted that might end up being fixed by the other comments

Also, it might be to do with how I'm viewing the page in my browser, but things like UUIDs are being truncated for me, would they need to be expanded to fit larger rendered strings?

In required metadata:
Screenshot 2024-06-25 at 14 51 51

In SDS metadata:
Screenshot 2024-06-25 at 14 49 10

static/javascript/launch.js
```diff
@@ -174,7 +178,7 @@ function setLaunchType(launchType) {
   if (launchType === "cir" || launchType === "remote" || launchType === "url") {
```
rmccar, line 178: This line could be made a bit simpler by doing something like:
if (["cir", "remote", "url"].includes(launchType)) {

rmccar:
> Really nice - great work! It looks loads better than before 😄 Just a few very minor cosmetic things I spotted that might end up being fixed by the other comments
> 
> Also, it might be to do with how I'm viewing the page in my browser, but things like UUIDs are being truncated for me, would they need to be expanded to fit larger rendered strings?
> 
> In required metadata: Screenshot 2024-06-25 at 14 51 51
> 
> In SDS metadata: Screenshot 2024-06-25 at 14 49 10

We have the space to make these a bit wider if we can think of another way of displaying the previous value button

templates/launch.html
```diff
@@ -1,215 +1,405 @@
...
+                            <span class="ons-btn__inner ons-accordion__toggle-all-inner"><span class="ons-btn__text">Show all</span></span>
+                        </button>
+                    </div>
+                    <div id="survey-type-metadata-accordion"
+                         class="ons-details ons-js-details ons-details--accordion ons-u-vh"
+                         data-group="metadata-accordion">
+                        <div class="ons-details__heading ons-js-details-heading"
+                             role="button"
+                             id="survey_type_metadata_detail">
+                            <h2 class="ons-details__title survey_heading"></h2>
```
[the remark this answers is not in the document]
VirajP1002: The reason it's initially empty is so that the business/health survey metadata titles can be dynamic when they need to be shown. The JS will insert the text when you load metadata via CIR (given the type you choose is business or health).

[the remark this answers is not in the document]
liamtoozer: Ah makes sense, thanks 👍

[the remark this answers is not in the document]
VirajP1002: I added some text within the h2 tags so it's easier to read, and this will be overwritten when the JS is applied 👍

static/javascript/launch.js
VirajP1002: Updated 👍

VirajP1002:
> > Really nice - great work! It looks loads better than before 😄 Just a few very minor cosmetic things I spotted that might end up being fixed by the other comments
> > Also, it might be to do with how I'm viewing the page in my browser, but things like UUIDs are being truncated for me, would they need to be expanded to fit larger rendered strings?
> > In required metadata: Screenshot 2024-06-25 at 14 51 51
> > In SDS metadata: Screenshot 2024-06-25 at 14 49 10
> 
> We have the space to make these a bit wider if we can think of another way of displaying the previous value button. Could try group the buttons with `ons-btn-group` like my other comment. They should wrap then but it would mean we would have a bigger gap between the inputs when they do.

Updated this to make the input boxes bigger, it involved some reshuffling of a few elements but it makes the result much cleaner

Screenshot 2024-06-26 at 15 44 02

static/javascript/launch.js
```diff
@@ -174,7 +178,7 @@ function setLaunchType(launchType) {
...
     if (schemaName.selectedIndex) {
       clearSurveyMetadataFields();
-      showSubmitFlushButtons(false);
+      enableSubmitFlushButtons(false);
```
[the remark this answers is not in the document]
VirajP1002: Removed and updated the param name 👍

static/javascript/launch.js
```diff
@@ -198,58 +202,59 @@ function setLaunchType(launchType) {
   }
 }
 
-function showSupplementaryData(show) {
+function showMetadataAccordion(type, show) {
+  let accordionElement = document.querySelector(
+    "#" + type + "-metadata-accordion",
```
liamtoozer, line 207: V minor/cosmetic - personal preference (so feel free to ignore!), but could use template literal / backticks here like `#${type}-metadata-accordion` like in the other places?

static/javascript/launch.js
```diff
@@ -621,12 +635,28 @@ function populateDropDownWithValue(selector, value) {
...
+  document.getElementById(metadataDetail).tabIndex = value;
+}
+
+function initialiseTabIndex() {
+  const details = [
+    "cir_metadata_detail",
+    "survey_type_metadata_detail",
+    "sds_metadata_detail",
+  ];
+  for (i = 0; i < 3; i++) {
```
liamtoozer, line 648: Minor - If the `details` array ever changes, the hardcoded `3` would also need to be changed. Could use something like `details.length` instead? Probably a super cool JS way of doing it but I'm not sure 😆

liamtoozer (approved):
Looks great to me! 👍

static/javascript/launch.js
VirajP1002: Updated 👍

static/javascript/launch.js
VirajP1002: Updated 👍

rmccar (approved):

Yuyuutsu (approved):
Looking nice 👍

PR #84, opened 2024-06-27
Update launcher to support multiple schema versions

VirajP1002:

Given the `schema_version` now has multiple versions, Launcher needed to be updated to support this where the data would update dynamically when choosing a different `sds_dataset_id`. Previously, when changing versions, it would add the dataset keys to the existing ones resulting in repeated fields. The changes made to the JS and HTML prevents that from occurring.

Ensure you use the Mock SDS [PR](https://github.com/ONSdigital/eq-runner-mock-sds/pull/19) when reviewing this to get the multiple schema versions for Supplementary Data. Make sure you use the Runner [PR](https://github.com/ONSdigital/eq-questionnaire-runner/pull/1414) too, to ensure that you can open schemas with multiple versions.

### How to review
- Ensure the changes make sense
- Ensure you can open schemas
- Ensure you select a different `sds_dataset_id` and the fields update correctly.

### DIFF
--- a/static/javascript/launch.js
+++ b/static/javascript/launch.js
@@ -142,6 +142,10 @@ let supplementaryDataSets = null;
 // We always need survey_id from top-level schema metadata for SDS retrieval
 let schemaSurveyId = null;
 
+const supplementaryDataSection = document.querySelector(
+    "#supplementary_data",
+);
+
 function clearSurveyMetadataFields() {
   document
     .querySelector("#survey-type-metadata-accordion")
@@ -398,10 +402,6 @@ function showCIRMetdata(cirInstrumentId, cirSchema) {
 function updateSDSDropdown() {
   const surveyId = schemaSurveyId;
   const periodId = document.getElementById("period_id")?.value;
-
-  const supplementaryDataSection = document.querySelector(
-    "#supplementary_data",
-  );
   const sdsDatasetIdElement = document.querySelector("#sds_dataset_id");
   loadSDSDatasetMetadata(surveyId, periodId)
     .then((sds_metadata_response) => {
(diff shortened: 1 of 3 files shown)


static/javascript/launch.js
```diff
@@ -543,6 +542,16 @@ function loadSupplementaryDataInfo() {
   }
 }
 
+function clearSDSMetadata() {
+  let dataset = document.querySelectorAll("#sds_data > div");
```
rmccar, line 546: I think it would be better to use something a bit more descriptive here than "div". Same with "div_array"

static/javascript/launch.js
```diff
@@ -543,6 +542,16 @@ function loadSupplementaryDataInfo() {
...
+  let div_array = [...dataset];
+  for (let i = 0; i < div_array.length; i++) {
+    if (div_array[i].classList.contains("sds-data")) {
```
rmccar, line 549: I think there should be a better way of working out if its the SDS data than a hardcoded class

static/javascript/launch.js
```diff
@@ -528,10 +526,11 @@ function loadSupplementaryDataInfo() {
     "sds_dataset_version",
   ];
 
-  const sdsMetadataSection = document.querySelector("#supplementary_data");
+  const sdsMetadataSection = document.querySelector("#sds_data");
   const sdsMetadataField = (key) =>
-    `<div class="ons-field ons-field--inline">${getLabelFor(key)}${getInputField(key, "text", selectedDataset[key], true)}</div>`;
+    `<div class="ons-field ons-field--inline sds_data">${getLabelFor(key)}${getInputField(key, "text", selectedDataset[key], true)}</div>`;
```
rmccar, line 531: I don't think we should be adding this `sds_data` class here to define if this is SDS data. We could use a data attribute or something instead

static/javascript/launch.js
```diff
@@ -400,14 +400,12 @@ function updateSDSDropdown() {
   const surveyId = schemaSurveyId;
   const periodId = document.getElementById("period_id")?.value;
 
-  const supplementaryDataSection = document.querySelector(
-    "#supplementary_data",
-  );
+  const supplementaryDataSection = document.querySelector("#sds_id");
```
rmccar, line 403: I think the id was more descriptive as it was. I dont think "sds_id" is a good id because it sort of describes what it is rather than setting it to an actual id that makes sense if Ive explained that right

static/javascript/launch.js
```diff
@@ -542,6 +543,16 @@ function loadSupplementaryDataInfo() {
...
 
+function clearSDSMetadata() {
+  let sds_values = document.querySelectorAll("#supplementary_data > div");
+  let sds_data = [...sds_values];
+  for (let i = 0; i < sds_data.length; i++) {
+    if (sds_data[i].hasAttribute("data-sds-metadata-key")) {
+      sds_data[i].remove();
+    }
+  }
+}
```
rmccar, line 554:
```suggestion
function clearSDSMetadata() {
  let sds_values = document.querySelectorAll("#supplementary_data > .ons-field");
  let sds_data = [...sds_values];
  for (let field of sds_data) {
    if (field.dataset.sdsMetadataKey) {
      field.remove();
    }
  }
}
```

Try this

rmccar (approved):

petechd (approved):

PR #86, opened 2024-07-10
Fix issue where the launch survey button being was disabled incorrectly

berroar:

The launch survey button being was disabled incorrectly when launching a supplementary data schema, then going back to launcher and opening a non supplementary data schema. This was due to an unnecessary call to the enableSubmitFlushButtons function when the handleNoSupplementaryData function was being called.

### How to review
Check that you are able to open a supplementary data schema, then launch a non supplementary data schema. Check no other scenarios are broken.


VirajP1002 (commented):
Tried with multiple scenarios and I cannot seem to recreate the issue with the new change 👍

liamtoozer (approved):
Works great for me 👍

VirajP1002 (approved):

PR #87, opened 2024-07-22
Fix mock SDS in test environment

rmccar:
Added if statement that checks for the test launcher url before calling the SDS service so that the error isn't caused in the test environment.

### How to review
Using the other 2 PRs related to this (https://github.com/ONSdigital/eq-terraform-gcp-launcher/pull/17/files https://github.com/ONSdigital/eq-pipelines/pull/612/files) which add a new environment variable to specify whether the SDS is available, spin up a copy of the test environment which includes the changes in these PRs
- Create a GCP project
- Create a pipeline that specifies the project id
In the pipeline var file:
- Update the slack webhooks to post to the "personal" channel
- Update the monitoring alerting channel to the "personal" channel
- Update the fqdn values
In the pipeline config file:
- Update the "questionnaire-launcher-image" to point to the "fix-mocksds-in-test" tag
- Update the "eq-terraform-gcp-launcher" repo to point to the "introduce-env-var-for-sds-unavailible" branch
Once the environment is spun up test launching some non SDS schemas to make sure they still launch correctly

Do the same for a staging environment which will have SDS available and test that SDS schemas can still be launched


petechd:
Incorrect base branch (this is raised against)?

rmccar:
> Incorrect base branch (this is raised against)?

Nah I raised it against Viraj's branch because it builds on an issue he fixed in that PR, so we need one more review on that one to get it merged in then this one can be changed to main

static/javascript/launch.js
```diff
@@ -497,12 +516,14 @@ function loadSchemaMetadata(schemaName, schemaUrl, cirInstrumentId) {
               })()}</div>`;
             })
             .join("");
-        updateSDSDropdown();
+        if (!window.location.href.includes("test-launcher.eq.gcp.onsdigital.uk")) {
```
berroar, line 519: Not sure we can just hardcode the test launcher url here - this probably needs to be configurable, as there may be other envs that could be impacted. For example, author raised an issue not being able to use launcher in their Preprod environment, as they don't have SDS running.

rmccar: Yeah I agree I thought there could be a better solution but as this was a dev tool I thought this would be good enough for now and could be changed later if we needed to add more envs to this list

README.md
```diff
@@ -92,26 +92,27 @@ https://golangci-lint.run/welcome/install/#local-installation to see additional
...
+| OIDC_TOKEN_BACKEND             | The backend to use when fetching the Open ID Connect token              | gcp                                                                    |
+| OIDC_TOKEN_VALIDITY_IN_SECONDS | The time in seconds an OIDC token is valid                              | 3600                                                                   |
+| OIDC_TOKEN_LEEWAY_IN_SECONDS   | The leeway to use when validating OIDC tokens                           | 300                                                                    |
+| SDS_OAUTH2_CLIENT_ID           | The OAuth2 Client ID used when setting up IAP on the SDS                |                                                                        |
+| CIR_OAUTH2_CLIENT_ID           | The OAuth2 Client ID used when setting up IAP on the CIR                |                                                                        |
+| SDS_UNAVAILABLE_IN_ENV         | If set signifies that the SDS service is unavailable in the environment |                                                                        |
```
berroar, line 118: I think this should be something like `SDS_ENABLED_IN_ENV`, as unavailable could mean it's down rather than it not being present? Also, at the moment it's almost like a double negative when it's set to false

rmccar: Yeah good point, Ill update that

VirajP1002:
Changes look good so far and it works functionally! 👍

Maybe it's worth putting the refactoring changes such as the README or the pyproject toml file in a separate PR as it might make it easier to review?

rmccar:
> Changes look good so far and it works functionally! 👍
> 
> Maybe it's worth putting the refactoring changes such as the README or the pyproject toml file in a separate PR as it might make it easier to review?

I could put the disabling of the "load metadata" button and refactoring of how buttons are enabled and disabled into another PR but the other changes are quite minor so I don't feel like its worth moving them to another PR

launch.go
```diff
@@ -116,19 +116,25 @@ func getSurveyDataHandler(w http.ResponseWriter, r *http.Request) {
 func getSupplementaryDataHandler(w http.ResponseWriter, r *http.Request) {
 	surveyId := r.URL.Query().Get("survey_id")
 	periodId := r.URL.Query().Get("period_id")
-
-	datasets, err := surveys.GetSupplementaryDataSets(surveyId, periodId)
-	if err != nil {
-		http.Error(w, fmt.Sprintf("GetSupplementaryDataSets err: %v", err), 500)
+	sdsEnabled := settings.Get("SDS_ENABLED_IN_ENV")
```
berroar, line 119: Need to document the new var in the readme 👍

static/javascript/launch.js
```diff
@@ -367,7 +367,7 @@ function getInputField(
 }
 
 async function loadSDSDatasetMetadata(survey_id, period_id) {
-  if (survey_id && period_id) {
+  if (survey_id && period_id && process.env.SDS_ENABLED_IN_ENV == "true") {
```
berroar, line 370: Should we use a strict equality operator here, for safety? And maybe assign the env var as a bool?

launch.go
```diff
@@ -116,19 +116,25 @@ func getSurveyDataHandler(w http.ResponseWriter, r *http.Request) {
...
+			return
+		}
+		datasetJSON, _ := json.Marshal(datasets)
+
+		_, writeError := w.Write([]byte(datasetJSON))
+		if writeError != nil {
+			http.Error(w, fmt.Sprintf("Write failed to write data as part of an HTTP reply: %v", writeError), 500)
+			return
+		}
+	} else {
```
liamtoozer, line 134: Minor - could maybe flip the logic here into a guard condition, so that we return from the function a bit earlier on if `sdsEnabled` is false?

rmccar: I've changed this to add a guard condition but because of the way its structured it should be exiting the function at a similar point. Have a look and let me know if this is what you meant.

launch.go
```diff
@@ -61,6 +61,7 @@ type page struct {
 	CirSchemas              []surveys.CIMetadata
 	AccountServiceURL       string
 	AccountServiceLogOutURL string
+	SdsEnabled              string
```
liamtoozer, line 64: This might need to be a `bool` type instead?

launch.go
liamtoozer: Looks good! 👍 Yep that's exactly what I had in mind

launch.go
rmccar: This cant be a bool because `settings.Get()` which is what is being used to get the env vars returns a string

liamtoozer: Yeah we're using standard library stuff to parse env vars as strings only atm :thinking: 

We could get around this by maybe having a struct to hold the different types? So like:
```
type Settings struct {
  StringSettings map[string]string
  BoolSettings map[string]bool
}
```
But then we would need to add different Getters and Setters for string type, and bool types.

I think we might want to avoid having string values for boolean values, though appreciate that this is a dev app, so maybe see what the others think? Using a Go package to do this sort of env parsing stuff for us might be useful, but might be out of scope for this work now?

rmccar: Yeah could be a good idea to have that discussion but I feel like anything we decide to do there is out of scope for this

static/javascript/launch.js
```diff
@@ -367,7 +366,7 @@ function getInputField(
 }
 
 async function loadSDSDatasetMetadata(survey_id, period_id) {
```
petechd, line 368: We could possibly pass in `sdsEnabled`(as sds_enabled) so we don't rely on correct scopes and  it would make it less error prone. Also we would then avoid using camel case and snake case on the same line so more consistency across the codebase.

liamtoozer (commented):
I just tried this locally but I'm unable to load SDS data via `test_supplementary_data` schema without manually changing `SDS_ENABLED_IN_ENV` to `true`. 

For dev convenience would it worth defaulting this to `true` so that we can just spin it up locally without changing, and then just overriding it with `false` for our relevant envs where needed?

static/javascript/launch.js
```diff
@@ -380,8 +382,8 @@ function getInputField(
   return `<input ${readOnly} id="${fieldName}" name="${fieldName}" type="${type}" ${value} class="ons-input ons-input--text ons-input--w-20" onchange="${onChangeCallback}">`;
 }
 
-async function loadSDSDatasetMetadata(survey_id, period_id) {
-  if (survey_id && period_id) {
+async function loadSDSDatasetMetadata(surveyId, periodId, sdsEnabled) {
+  if (surveyId && periodId && sdsEnabled) {
     const sds_dataset_metadata_url = `/supplementary-data?survey_id=${survey_id}&period_id=${period_id}`;
```
petechd, line 387: These two need changing now, otherwise it won't work.

templates/launch.html
```diff
@@ -392,6 +392,7 @@ <h2 class="ons-details__title">Runner Metadata</h2>
         </div>
     </form>
     <script src="https://cdn.ons.gov.uk/sdc/design-system/70.0.7/scripts/main.js"></script>
+    <script>const sdsEnabled = {{ .SdsEnabled }} === "true"</script>
```
petechd, line 395: This is getting highlighted by the linters, can we pass the Go var in quotes like in other places in this file and cast it as a bool instead of comparison?

rmccar: Which linters? I cant see anything on the lint action or anything in my IDE being highlighted

petechd (approved):

PR #90, opened 2024-08-12
Disable load metadata button when fields aren't passing validation

rmccar:
This PR refactors the way buttons are enabled and disabled and updates launcher to disable the "load metadata" button when the correct fields aren't selected

### How to review
- Test that the "load metadata" button is only enabled when the metadata fields are selected and/or input
- Test that it is then disabled again when local storage is cleared
- Test that it is enabled and disabled when the metadata fields are selected and/or input and then cleared again


VirajP1002 (commented):
Minor - however, when following the steps:
- I loaded metadata by `Remote Schemas` where I chose the survey type and schema then the button was enabled
- Then I loaded metadata by `Schema Name` and this disabled the `Load Metadata` button as expected
- But then I loaded metadata by `Remote Schema` again, and I only added the survey type and **not** the schema name and the button was enabled which I'm guessing isn't correct

Happy to jump in a huddle if that doesn't make sense

rmccar:
> Minor - however, when following the steps:
> 
> * I loaded metadata by `Remote Schemas` where I chose the survey type and schema then the button was enabled
> * Then I loaded metadata by `Schema Name` and this disabled the `Load Metadata` button as expected
> * But then I loaded metadata by `Remote Schema` again, and I only added the survey type and **not** the schema name and the button was enabled which I'm guessing isn't correct
> 
> Happy to jump in a huddle if that doesn't make sense

Should be fixed now if you want to give it another test

static/javascript/launch.js
```diff
@@ -152,54 +165,72 @@ function clearSurveyMetadataFields() {
...
     cirSchemas.selectedIndex = 0;
     remoteSchemaSurveyType.selectedIndex = 0;
+    cirSchema = null;
+    surveyType = null;
+    schemaUrl = null;
     localStorage.removeItem("schema_url");
     localStorage.removeItem("cir_schema");
     localStorage.removeItem("survey_type");
     document.querySelector("#language_code").disabled = false;
+    enableDisableButtons([loadMetadataButton], false);
```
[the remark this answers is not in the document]
petechd: `enableDisableButtons()` used to do some extra work now it just passes the flag further but nothing else.

static/javascript/launch.js
```diff
@@ -152,54 +165,72 @@ function clearSurveyMetadataFields() {
   showMetadataAccordion("sds", false);
 }
 
-function setSurveyType(event) {
-  localStorage.setItem("survey_type", event.value);
-  setLaunchType("remote");
+function validateRemoteSection() {
```
petechd, line 168: Maybe we should focus on what it does exactly so something more like  `setLoadMetadataButtons()`?

rmccar: To me `setLoadMetadataButtons` isn't that descriptive of what the function actually does, Ive updated it to `enableDisableLoadMetadataButton` although I do think this is quite a long function name

petechd: Maybe we should use `toggle`? Sort of confusing since enable/disable are opposites.

static/javascript/launch.js
```diff
@@ -152,54 +165,72 @@ function clearSurveyMetadataFields() {
...
   console.log(schemaName);
-  const remoteSchemaSurveyType = document.querySelector(
-    "#remote-schema-survey-type",
-  );
 
   if (["cir", "remote", "url"].includes(launchType)) {
     if (schemaName.selectedIndex) {
       clearSurveyMetadataFields();
-      enableSubmitFlushButtons(false);
+      disableButtons(launchFlushButtons);
```
petechd, line 206: Could we, for clarity, make it an array and:
```js
disableButtons([launchButton, flushButton]);
```
and in the other place:
```js
const launchButton = document.querySelector("#launch-btn");
const flushButton =  document.querySelector("#flush-btn");
];
```
?
Otherwise it might confuse someone into thinking it's a verb and we are trying to launch flush buttons.

petechd: Otherwise the changes look fine.

petechd (approved):

liamtoozer (approved):
Works great for me 👍

PR #93, opened 2025-01-21
Fix: Correct DOM Text Handling, Pin Tags in Workflow, and Add Permissions

Yuyuutsu:
Code Scanning has been enabled in some of our Github Repositories is highlighting issues. 
This PR is to fix these issues:
- DOM text reinterpreted as HTML
- Workflow does not contain permissions
- Unpinned tag for a non-immutable Action in workflow

### How to review
Check if the changes make sense 
Test Launcher locally to see if the HTML changes do not affect Launcher Visually

### DIFF
--- a/static/javascript/launch.js
+++ b/static/javascript/launch.js
@@ -272,19 +272,19 @@ function includeSurveyMetadataFields(schema_name, survey_type) {
     .classList.remove("ons-u-vh");
   document.querySelector(".survey_heading").innerHTML =
     `${survey_type} Survey Metadata`;
-    
+
   const surveyMetadataFields = document.querySelector(
-      "#survey_metadata_fields",
-    );
-    const div = document.createElement("div");
-    div.className = "ons-field ons-field--inline";
-    div.innerHTML = `
+    "#survey_metadata_fields",
+  );
+  const div = document.createElement("div");
+  div.className = "ons-field ons-field--inline";
+  div.innerHTML = `
     <label class="ons-label" for="form_type">form_type</label> 
     <input id="form_type" name="form_type" type="text" class="ons-input ons-input--text ons-input-type__input">
     `;
   div.querySelector("input").value = formTypeValue;
-    surveyMetadataFields.textContent = '';
-    surveyMetadataFields.appendChild(div);
+  surveyMetadataFields.textContent = "";
+  surveyMetadataFields.appendChild(div);
   setTabIndex("survey_type_metadata_detail", 0);
   showMetadataAccordion("sds", true);
 }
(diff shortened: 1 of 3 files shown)


static/javascript/launch.js
```diff
@@ -1,4 +1,13 @@
 // uuidv4: from https://github.com/kelektiv/node-uuid
+function escapeHtml(unsafe) {
```
berroar, line 2: Is this actually used anywhere? How come this has been added?

Yuyuutsu: Oops forgot to apply the escapeHTML function. Added now.

Yuyuutsu: It was a suggested change so "any text taken from the DOM and used to set innerHTML is properly escaped to prevent XSS attacks".

static/javascript/launch.js
```diff
@@ -262,13 +271,20 @@ function includeSurveyMetadataFields(schema_name, survey_type) {
...
   document.querySelector(".survey_heading").innerHTML =
-    `${survey_type} Survey Metadata`;
+    `${escapeHtml(survey_type)} Survey Metadata`;
 
-  document.querySelector("#survey_metadata_fields").innerHTML =
-    `<div class="ons-field ons-field--inline">
-      <label class="ons-label" for="form_type">form_type</label>
-      <input id="form_type" name="form_type" type="text" value="${formTypeValue}" class="ons-input ons-input--text ons-input-type__input">
-    </div>`;
+  const surveyMetadataFields = document.querySelector(
```
berroar, line 276: According to the alert this is one of many instances where we will need a fix?

berroar (changes requested):

static/javascript/launch.js
Yuyuutsu: yeah there are two alerts in this file. This change is suggested from here:
https://github.com/ONSdigital/eq-questionnaire-launcher/security/code-scanning/8

petechd: There are actually 18 changes suggested when you press "show paths"?

Yuyuutsu: I have had a look. I think it is just showing the path leading to the error. As you can see at the end of the path is what the codeQL suggests fixing.

petechd: Ok, so fixing the issue from the last step in both alerts will clear them?

petechd (approved):

berroar (approved):

static/javascript/launch.js
Yuyuutsu: Yeah, thats what it looks like. as only the parts i have changed have dom text issues

PR #96, opened 2025-04-14
Upgrade Poetry to v2.1.2

petechd:
This PR upgrades Poetry to `v2.1.2`. This involved regenerating the lock file and updating poetry versions across the repo to use this version.

### How to review
- Ensure all Make commands work as before
- Ensure tests & linting pass
- Ensure the Dockerfile builds correctly.
- Ensure there's no warning in Actions


Yuyuutsu (commented):
I am getting go linting errors. However I am on `golangci-lint version 2.1.6 built with go1.24.2`.This may not be in scope but we may want to upgrade our go and golangci-lint versions and resolve the linting errors.

petechd:
> I am getting go linting errors. However I am on `golangci-lint version 2.1.6 built with go1.24.2`.This may not be in scope but we may want to upgrade our go and golangci-lint versions and resolve the linting errors.

I would guess this is part of another task, since we want to update `poetry` only here, but we can ask other devs for opinions.

Yuyuutsu:
> > I am getting go linting errors. However I am on `golangci-lint version 2.1.6 built with go1.24.2`.This may not be in scope but we may want to upgrade our go and golangci-lint versions and resolve the linting errors.
> 
> I would guess this is part of another task, since we want to update `poetry` only here, but we can ask other devs for opinions.

Yeah happy to approve since this PR is specifically for Poetry

Yuyuutsu (approved):

PR #98, opened 2025-05-20
Create codeql-analysis.yml

Yuyuutsu:
We want to be consistent in all eQ repositories.
This PR is to change our codeQL settings from the default to advanced to align with Runner

### How to review
Check CodeQL YML settings make sense
CodeQL Scan runs as expected


.github/workflows/codeql-analysis.yml
```diff
@@ -0,0 +1,45 @@
+name: "CodeQL Advanced"
+
+on:
+  push:
+    branches: [ "main", "v3-census" ]
```
liamtoozer, line 5: Is `v3-census` in use?

liamtoozer (approved):

